PR #51, opened 2015-08-04
Converted 'make_version' script to python

### LINKED ISSUE #50: Compilation error when running `make_version` during build process on OS X
I am trying to build VST (in conjunction with reading Programming Logics for Certified Compilers) on the latest version OS X (10.10) and I get the following error when running `make_version`:
```
...
sh util/make_version
date: illegal option -- -
usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
COQC version.v
File ".../VST-1.5.3/version.v", line 2, characters 0-1:
Error: No focused proof (No proof-editing in progress).
make: *** [version.vo] Error 1
```

This happens both on the latest stable release and the git master branch.
I believe this is due to the `make_version` script being dependent on the GNU versions of the `date` and `echo` commands.

I think there might be multiple ways to mitigate this problem:
* Making a note in the _BUILD_ORGANIZATION_ file stating that one needs to install _coreutils_  using MacPorts or Homebrew and somehow replace the `date` and `echo` calls with `gdate` and `gecho` calls (or link the `date` and `echo` commands to the custom version)
* or Making a portable sh or python script that is not dependent on underlying OS as the `make_version` script is now

I am willing to provide a PR if there is any solution preferred and no one else makes one.

ahmadsalim:
This is done in order to ensure that 'VST' builds across platforms correctly and not depend on GNU versions of command line utilities which may not be available on other platforms.

Fixes #50.


andrew-appel:
I fixed this in a different way, in commit fa8004fc7530eab09bf468be9bf7f9128f1eedd8.

PR #55, opened 2016-04-26
make forward_while also work if the while loop is at the end of a block

samuelgruetter:
Hi,

when I wanted to prove a specification for a function whose body contains a block ending with a while loop, i.e. something of the form

```
if (...) {
  ...
  while (...) {...}
}
```

I could not apply the `forward_while` tactic (after `forward`ing until just before the while).
See `progs/verif_while_at_end_of_block.v` and `progs/while_at_end_of_block.c` for a small example exhibiting the problem.
In tactic `forward_while` in file `floyd/forward.v`, the line `eapply semax_seq;` fails, because it expects a goal of the form `semax _ _ (Ssequence (Swhile _ _) _) _`, whereas our goal is of the form `semax _ _ (Swhile _ _) _`.

I considered two ways to fix this:
-    Not applying `semax_seq` if we don't have an `Ssequence`. But then, nor `semax_while_3g1` nor `semax_while` can be applied directly, and the two cases (ie whether there is an `Ssequence` around the `Swhile` or not) have to be treated separately (or rather, seem tricky to be handled together ;-) )
-    Making the goal a bit more complex by turning the `(Swhile _ _)` into a `(Ssequence (Swhile _ _) Sskip)`, just to undo this step in the next line. This sounds a bit silly, but allows a uniform treatment of both cases.

The second approach is in this PR, I'd love to hear what you think about it ;-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,7 +187,7 @@ PROGS_FILES= \
   verif_nest3.v verif_nest2.v \
   logical_compare.v verif_logical_compare.v field_loadstore.v  verif_field_loadstore.v \
   even.v verif_even.v odd.v verif_odd.v \
-  merge.v verif_merge.v
+  merge.v verif_merge.v while_at_end_of_block.v verif_while_at_end_of_block.v
 # verif_message.v verif_dotprod.v verif_insertion_sort.v 
 
 SHA_FILES= \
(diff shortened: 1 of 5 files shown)


andrew-appel:
Reimplemented, no need to pull this now.

PR #58, opened 2016-06-18
Allow overriding of the COMPCERT variable

tbrk:
Declaring the value of COMPCERT with ?= allows it to be overridden from the
command-line, for example,

    make -C VST COMPCERT=../CompCert

(refer: https://www.gnu.org/software/make/manual/make.html#Setting-Variables)

This is useful, in particular, for including both VST and CompCert as git
submodules of another project without having to modify (and
commit/branch/share/etc.) their contents.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # See the file BUILD_ORGANIZATION for 
 # explanations of why this is the way it is
 
-COMPCERT=compcert
+COMPCERT ?= compcert
 -include CONFIGURE
 #Note:  You can make a CONFIGURE file with the definition
 #   COMPCERT=../compcert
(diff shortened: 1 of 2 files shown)


tbrk:
I have just started looking at using the VST as a submodule of another project that also involves CompCert. The CompCert load paths all begin with "compcert", e.g., compcert.lib, compcert.common, etc., whereas the VST ones are unadorned, e.g., msl, sepcomp, etc. Would it be worth adding a prefix to their namespace? Should I raise an issue?

andrew-appel:
I merged this pull-request "by hand."  Thanks.

andrew-appel:
Yes, go ahead and raise an issue regarding your comment about load paths.

tbrk:
Great. Thank you.

PR #61, opened 2017-01-19
add the possibility to compile a html doc in doc/html

ildyria:


andrew-appel:
The Travis-CI failure is not caused by this pull request, it's a timeout in the mailboxes proof that's being worked on independently.

PR #65, opened 2017-01-24
add proof of factorial intro progs

ildyria:


andrew-appel:
Thank you for this example program, but I don't think it's needed in our distribution.

PR #67, opened 2017-03-22
Explicitly name parameter in destruct.

ejgallego:
This is needed in order for VST to compile with Coq 8.7.

The usual annoyance with automatically generated names, in this case the type changes, and so does the name. This patch thus makes the destruct more robust.

See https://github.com/coq/coq/pull/489 for more context.


maximedenes:
Hello,

Could we get this merged? It is a trivial, backward compatible fix that would enable us (the Coq development team) to continuously test the compilation VST, making future transitions to the next Coq versions smoother for you.

PR #76, opened 2017-07-05
improve 'make clean' to also remove .xxx.vo.aux

ildyria:
`make clean` was forgetting some files such as:
- `version.vo`,
- `version.glob`,
- `.version.vo.aux`
- all the `.xxx.vo.aux`
- `_CoqProject` (as it is also generated)
- `.lia.cache` and `.nia.cache`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -375,6 +375,26 @@ FILES = \
  $(HMACDRBG_Files:%=hmacdrbg/%)
 # $(DRBG_FILES:%=verifiedDrbg/spec/%)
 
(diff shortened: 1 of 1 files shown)


ildyria:
Check failed because: `The job exceeded the maximum time limit for jobs, and has been terminated.`

mansky1:
This happens sometimes, but Travis seems to cache the results of some steps: I'll try running the test again.

mansky1:
It got there eventually. :)

PR #82, opened 2017-07-31
Improve dep

QinxiangCao:
Improve some lemmas like field_address_app and field_address_isptr: some unnecessary assumptions are removed.

Redefine denote_efield. Now we use "local (denote_efield efs gfs)" instead of "denote_efield efs gfs". So, it is a pure fact by shape.


QinxiangCao:
Ltac computable msubst_efield_denote is also added.

PR #87, opened 2017-08-11
Made with_library much faster, especially the version in floyd/library.v

andrew-appel:


andrew-appel:
I tested everything fully on my own machine, so the travis tests are redundant; I'm merging it now.

PR #89, opened 2017-08-13
Small change for go_lower

QinxiangCao:
Previously, the intro tactic for the local clauses in go_lower call "trivial" lemma first if the right side is independent with environ. This causes some incoherent behavior when advanced technique can simplify the right side and make it independent with environ.


QinxiangCao:
Now, trivial lemma is removed from go_lower tactic. No efficiency gain or lost is observed.

PR #91, opened 2017-08-15
Canon semax return

QinxiangCao:
Add lemmas in canonical form to handle return command. Now, the entailment that left to users are directly generated by those lemmas while it was originally generated by rewrites.

In efficiency improvement of 8%-40% (15% at average) is observed in the new forward_return tactic.


QinxiangCao:
This change has one small drawback---postconditions must be written in existential canonical form (except main_post). Error messages are added for this.

PR #116, opened 2017-11-16
Remove uses of deprecated appcontext.

Zimmi48:
`appcontext` is likely to be removed in Coq 8.8 and is semantically equivalent to `context`


Zimmi48:
Thanks!

PR #129, opened 2018-01-01
Align

QinxiangCao:
This update is to solve the problem that previous legal_alignas is too strong so that some C standard library programs are not provable. The cause of this problem is a mismatch between top level alignof setting and low level alignment requirement by hardware read/write.

Now, we do not require alignof and Memdata.align_chunk to coincide. Instead, we only require the addresses (which is computed by top level layout based on alignof) satisfies hardware read/write requirement (which is computed by Memdata.align_chunk). Thus align_compatible is redefined.

This cause a potential problem that: "data_at tuint" and "data_at tint" are no longer beta-eta equivalent;  "data_at tint" of two different compspecs are no longer beta-eta equivalent. Now, change_compspecs are done by rewrite instead of change.

Further test is needed to evaluate efficient gain/lose of this change about change_compspecs. It is know for now that the original method will be super slow on many-field structs.


QinxiangCao:
Here is several test cases:

Test 1:

New forward_call:
      4.593 secs (2.312u,0.093s)
      3.622 secs (1.578u,0.015s)
      3.675 secs (1.515u,0.031s)
      3.551 secs (1.406u,0.062s)
      3.557 secs (1.484u,0.015s)
      3.507 secs (1.375u,0.015s)

Old forward_call:
      5.584 secs (1.484u,0.015s)
      5.643 secs (1.781u,0.s)
      5.586 secs (1.531u,0.015s)
      5.478 secs (1.625u,0.015s)
      5.476 secs (1.437u,0.s)


Test 2:

New forward_call:
     9.441 secs (2.078u,0.046s)
     18.945 secs (1.046u,0.s)
     19.08 secs (1.187u,0.031s)
     14.854 secs (1.25u,0.015s)
     19.352 secs (1.25u,0.s)

Old forward_call:
     19.707 secs (1.484u,0.015s)
     18.621 secs (1.062u,0.s)
     18.669 secs (0.968u,0.031s)
     19.347 secs (1.406u,0.015s)
     18.725 secs (1.u,0.s)

So, a brief conclusion is that this new forward_call is not making things worse.

PR #140, opened 2018-01-18
Don't depend on the judgmental behavior of Bool.eqb (version 2)

JasonGross:
This change is backwards compatible, and makes flocq compatible with https://github.com/coq/coq/pull/6596.

This has now been merged into both flocq and CompCert at https://gitlab.inria.fr/flocq/flocq/commit/db356aa5ea0fd0234e3872f70e8972086055cd57 and https://github.com/AbsInt/CompCert/pull/217.


andrew-appel:
Now that we are after the VST 2.0 release, and don't require hypercompatibility with CompCert 3.2 -- instead we are happy to have compatibility with the CompCert trunk -- I am happy to merge this pull request.

PR #149, opened 2018-02-19
Improve loadstore

QinxiangCao:
Add a new general tactic for finding nth element in SEP clauses satisfying condition. This replaces original find_type_conflict and sc_new_instantiate.

This enables some finer grained manipulation. Now, a forward for a store command will not turn a data_at into field_at nil.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ env:
 branches:
   only:
   - master
-  - align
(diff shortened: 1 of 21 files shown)


QinxiangCao:
Now, load/store lemmas and tactics are also cleaned up.

PR #164, opened 2018-03-28
Do not disable Coq warnings

maximedenes:
Warnings that are on by default require user action. They should be addressed, not ignored. For example, deprecation warnings make it easier to port developments smoothly to new versions of Coq.


andrew-appel:
This pull request is implemented by commit ccc1d58.

maximedenes:
Thanks!

PR #167, opened 2018-03-29
Avoid relying on bug Coq bug #6998

maximedenes:
This patch may need to be refined, since it probably reveals a problem
with some abstraction barriers like mpred.

But at least, it will make it possible for Coq's simple apply to respect
Opaque without breaking VST.

### DIFF
--- a/concurrency/semax_conc_pred.v
+++ b/concurrency/semax_conc_pred.v
@@ -149,7 +149,7 @@ Proof.
   - rewrite prop_true_andp; auto.
   - rewrite prop_false_andp; auto.
     apply pred_ext.
-    + unfold lock_inv. Intros b ofs. subst; simpl in *; tauto.
+    + unfold lock_inv. Transparent mpred. Intros b ofs. Opaque mpred. subst; simpl in *; tauto.
     + apply FF_left.
 Qed.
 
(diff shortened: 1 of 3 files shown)


andrew-appel:
Merged manually.  Instead of using "Transparent. auto. Opaque.", just use "apply derives_refl" instead.

maximedenes:
> Merged manually. Instead of using "Transparent. auto. Opaque.", just use "apply derives_refl" instead.

Hi Andrew,

I believe this is not a good fix, because it relies on a similar bug for `apply`. So when I'll fix `apply`, VST will break again...

andrew-appel:
If you change "apply", then so many other things will break in VST that I will have to make the changes, and this one will be the least of our problems.

maximedenes:
I see. What is your take on the general problem? Is the fact that VST has to bypass the `Opaque` flag a problem with VST or a flaw in the design on the Coq side (like the `Opaque` flag being too global)?

andrew-appel:
Perhaps it is a design flaw in Coq that "apply" bypasses Opaque, but that does not cause me any severe problems.  The severe problem is that even "simple apply" bypasses Opaque.

In the future it would be worth doing some experiments to see how many different developments break if you change the behavior of "apply", but I don't consider this to be urgent.  

This is a design discussion that might be best conducted in a venue other than the comments section on github.

PR #171, opened 2018-04-05
Improve forward_call

QinxiangCao:
Use find_nth_SEP to improve the cancel inside forward_call. The efficiency is 5% improved
[Improve forward_call.xlsx](https://github.com/PrincetonUniversity/VST/files/1882219/Improve.forward_call.xlsx)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ env:
 branches:
   only:
   - master
-  - nth_SEP
 notifications:
   email:
     recipients:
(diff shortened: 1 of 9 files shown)


QinxiangCao:
The previous test does not show the correct number of difference since cancel is not the bottleneck of forward_call. The attachment can be used to test "cancel" alone inside a forward_call. To measure time more accurately, it does cancel for 10 same goals. Syntactic_cancel is 7 times faster on this case.

Copy it to line 389 of progs/verif_bst.v  when using it for testing.
[test.txt](https://github.com/PrincetonUniversity/VST/files/1885432/test.txt)

PR #199, opened 2018-04-29
Test new cancel

QinxiangCao:
Use new symbolic cancel tactic to replace the original cancel.

The efficiency is improved in the following way:
1. In most simple cases, it cut down from around 0.2s to 0.015s. (* Tested by running 700 cancel together. *)
2. In the cases that 3-4 simple data_at/field_at on each side, the efficiency is slight improved (* less than 10% *)
3. In the cases that 12 data_at/field_at on each side and the reptype is large, it cuts down from 700+ seconds to 22 seconds. (See aes/verif_gen_tables_LL.v line 519. *)


QinxiangCao:
The test for aes/verif_gen_tables_LL.v line 519:

Syntactic cancel:                          25.504 22.444 22.087
Original cancel after fixing forward ret:  forever
Original cancel before fixing forward ret: 757.983 762.951 812.217.

QinxiangCao:
By testing cancel with 24 simple sep-conjunct on each side, cut down from 1.64s to 0.11s (by running cancel 80 times).

PR #225, opened 2018-05-31
Fix tactic level mistake about (assert_PROP + by) etc.

QinxiangCao:
Luke Chen find that "assert_PROP (P) by tac1; tac2" was interpreted as "assert_PROP (P) by (tac1; tac2)".


QinxiangCao:
Fix tactic level mistake about (assert_PROP + by) etc. Qinshi Wang found that "assert_PROP (P) by tac1; tac2" was interpreted as "assert_PROP (P) by (tac1; tac2)".
Now, it is fixed, i.e. "assert_PROP (P) by tac1; tac2" is now interpreted as "(assert_PROP (P) by tac1); tac2". (Previously, I remembered the wrong name.)

PR #242, opened 2018-06-26
Roconnor blockstream reptype

QinxiangCao:
Now, reptype (tarray t _) is reducible to (list (reptype t)). Previously they are only provably equal.

Test whether examples work efficiently.


QinxiangCao:
After fixing an efficiency bug:
33'06'' + 22'32'' + 15'51''

Issue #218, opened 2018-05-24
forward_for_simple_bound cannot tolerate return statement


andrew-appel:
If the body of a for loop contains a return statement, the loop cannot be verified using forward_for_simple_bound.   It should be possible to remove this restriction.

Issue #193, opened 2018-04-24
Forward_for and forward_for_simple_bound cannot handle multiple EX in Inv


QinxiangCao:
Forward_for and forward_for_simple_bound cannot handle multiple existential quantifiers in loop invariants.

QinxiangCao:
Forward_for_simple_bound now allows multiple EX. (PR #260)

PR #260, opened 2018-07-16
New forward_for_simple_bound

QinxiangCao:
This PR follows PR #253. A new forward_for_simple_bound tactic is built; the proof rule on which the tactic is based is fully debugged now.

I improve the tactic in the following aspect:

(1) It allow continue and return in loop body.
(2) It allow multiple EX in loop invariant.
(3) In the beginning of loop body verification, the existentially quantified variable in the loop invariant will not always be extracted as "x". The original name used in the invariant expression will be used.


QinxiangCao:
Half of issue #193 is solve. Fix issue #218. Close issue #218.

PR #276, opened 2018-08-19
Changes in Makefile for better compability with external CompCert 

andrew-appel:
(i.e., method B in BUILD_ORGANIZATION)
In particular, the -R flags to coqc are more selective of which subdirectories of compcert they import.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ compcert/clightgen
 compcert/clightgen.byte
 compcert/tools/ndfun
 compcert/tools/modorder
-compcert/Makefile.config
 # Generated files
 compcert/.depend.extr
 compcert/compcert.ini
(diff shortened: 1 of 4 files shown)


Zimmi48:
This broke our VST CI tests (both on master and v8.8) but I can't really explain why. Here is a log: https://gitlab.com/coq/coq/-/jobs/90883746
The error:
```
File "./compcert/exportclight/Clightdefs.v", line 19, characters 29-37:
Error: Unable to locate library Integers with prefix compcert.
```

When trying to reproduce the problem locally (with Coq 8.8.1), I don't get the same error, I get:
```
File "./compcert/x86_32/Archi.v", line 20, characters 15-26:
Error: Unable to locate library Fappli_IEEE.
```
instead!

Note: we run `make` with `IGNORECOQVERSION=true`.

Sorry for not being more precise. I haven't investigated this much.

PR #285, opened 2018-09-05
Add several lemmas about weak_valid_pointer.

txyyss:
In proving a program using VST, when it goes through a pointer comparison like "if (s < t)", it requires that s and t should both be weak_valid_pointer and in the same block. In my case, it is possible that s or t is in the end of an array. Thus, I can not apply the following lemma:

Lemma valid_pointer_weak: forall a : val, valid_pointer a |-- weak_valid_pointer a.

because it is not a valid_pointer. So I proved a new one for my case:

Lemma memory_block_weak_valid_pointer: forall {cs: compspecs} sh n p i,
  0 <= i <= n -> 0 < n -> nonidentity sh ->
  memory_block sh n p |-- weak_valid_pointer (offset_val i p).

When I sep_apply this entailment, I found that I also need the following lemma to finish the whole thing:

Lemma extend_weak_valid_pointer:
  forall p Q, weak_valid_pointer p * Q |-- weak_valid_pointer p.

I think they are also helpful for other users.

Thank you,

Shengyi Wang

### DIFF
--- a/floyd/entailer.v
+++ b/floyd/entailer.v
@@ -290,6 +290,22 @@ intros.
  hnf; eauto.
 Qed.
 
+Lemma extend_weak_valid_pointer:
+  forall p Q, weak_valid_pointer p * Q |-- weak_valid_pointer p.
+Proof.
+  intros. unfold weak_valid_pointer.
+  pose proof (extend_tc.extend_valid_pointer' p 0).
+  pose proof (predicates_hered.boxy_e _ _ H). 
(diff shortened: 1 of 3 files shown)


txyyss:
The CI failed for some reason I don't understand. From log, it is an "apt-get install failed"

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $TRAVIS_APT_OPTS install opam aspcud" failed and exited with 100 during .

PR #287, opened 2018-09-15
Improve ret_assert related stuff

QinxiangCao:
Simplify the definition of proj_ret_assert to make later porting convenient.


QinxiangCao:
1. Simplify the definition of proj_ret_assert to make later porting convenient.
2. Unify guard and rguard by _guard.
3. Add one more consequence rule for fun_id (both in pre and post, we did have pre case only).

Issue #268, opened 2018-07-22
Unroll a for loop


txyyss:
During the proving process, could VST unroll a for loop so as to reduce the proof burden?

For example, in line 142 of [gc.c](https://github.com/PrincetonUniversity/certicoq/blob/master/theories/Runtime/gc.c#L142), the initial value of i is -1, which is a special position of an object block. The predicate about this position is different from others. So if I can unroll the -1 loop, the invariant of the loop will be much clearer.

PR #289, opened 2018-09-21
Separation logic as logic

QinxiangCao:


QinxiangCao:
The separation Hoare logic, which used in VST/floyd, was shallowly embedded. Now, the logic is deep embedded (despite the fact that the assertion language is still shallowly embedded) --- see floyd/SeparationLogicAsLogic.v and the Module AuxDefs in it.

Why doing that? For more convenient derivation of "semax_if_seq" and "semax_loop_nocontinue". Now, they are proved as meta-theorems of Clight Separation Logic, i.e. we can prove that by induction over the Hoare logic proof tree. Original proof of semax_if_seq is long. And semax_loop_nocontinue is even too difficult to get proved.

Now, seq_assoc, seq_skip, skip_seq are also meta-theorems. Even extraction rules (semax_extract_exists, semax_extract_prop, semax_extract_later_prop) and the frame rule are meta-theorems. This cut down 1500+ lines of proofs.

Now, besides the interface in veric/SeparationLogic and veric/SeparationLogicSoundness, we provide another interface call MainTheorem in floyd/SeparationLogicAsLogicSoundness.v.

A careful choice of the proof rules in Minimum Separation Logic is critical (see veric/SeparationLogic.v) because that will decide which proof trees will be allowed. Rebuilding semax_extract_exists as a meta-theorem is hard (see floyd/SeparationLogicAsLogic.v).

Minor stuff: #268 is solved by providing semax_loop_unroll1.

QinxiangCao:
VST correctly builds on my laptop. But the 5th travis task always fails.

mansky1:
Weirdly, it seems to be having trouble loading a compiled file. Maybe clearing the cache and rebuilding would fix it?

PR #290, opened 2018-09-24
Use Defined for Instances instead of Qed

jbaum98:
Using `Qed` for typeclass instances creates opaque identifiers that can't be evaluated during computation.
```coq
Compute 1 ?= 1.
     = (let (eqb, _) := nat_EqDec in eqb) 1 1
```
If we use `Defined` instead, the definitions are transparent and can be evaluated as expected.
```coq
Compute 1 ?= 1.
     = true
```

### DIFF
--- a/fcf/EqDec.v
+++ b/fcf/EqDec.v
@@ -189,7 +189,7 @@ Section Vector_EqDec.
   subst.
   apply eqb_vector_refl.
   
-  Qed.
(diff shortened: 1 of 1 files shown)


andrew-appel:
Closing this pull request without merging, because it would be better submitted to the fcf repo, https://github.com/adampetcher/fcf

PR #303, opened 2018-12-20
Compatibility for coq/coq#8817 (SProp)

SkySkimmer:
Coq now detects that the R and R_res arguments to
find_nth_preds_constr shouldn't count to its level as they're wholly
determined from the type. This causes generation of the _rec induction
principle whose name is already used.

Same for construct_fold_right_sepcon.

This is backward compatible so should be fine to merge.


SkySkimmer:
In the end I removed the Coq side so this won't be needed any more. The patch may still have value, and maybe the improvement will come back some other time, but for now merge or close as you prefer.

PR #304, opened 2018-12-23
Do not rely on `Refine Instance Mode`

maximedenes:
This option is soon going to be turned off by default. See
https://github.com/coq/coq/pull/9270

This patch should be backward compatible (but that should be tested).


maximedenes:
Let's wait to see what we do in https://github.com/AbsInt/CompCert/pull/261

maximedenes:
The failure seems unrelated. Could this PR be merged? Thanks!

maximedenes:
Thanks!

PR #305, opened 2019-01-07
Fix a few more occurrences of `Instance`

maximedenes:
This is a follow-up on #304, where I had missed a few occurrences of `Instance` relying on `Set Refine Instance Mode`. Apologies for the noise.

### DIFF
--- a/progs/ghosts.v
+++ b/progs/ghosts.v
@@ -97,11 +97,8 @@ Qed.
 
 End ghost.
 
-Instance exclusive_PCM A : Ghost := { valid a := True;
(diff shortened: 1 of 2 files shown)


maximedenes:
@andrew-appel would you mind merging this one too, so that we can proceed with https://github.com/coq/coq/pull/9270 on our side? Thank you and sorry again for having missed these occurrences initially.

andrew-appel:
I had been waiting to merge this because the travis-ci check failed, but the failure seems to have nothing to do with this pull request, it's a travis opam problem.  So I will merge this and hope for the best.

PR #309, opened 2019-01-26
add forward_call that succeeds only if full WITH-clause inferred

bcip:


bcip:
@andrew-appel Check failed because opam does not have coq 8.9.0.

PR #311, opened 2019-02-12
Remove spurious `Proof.` in text.

ejgallego:
The command will require an actual open proof in 8.10

This is backwards compatible.


ejgallego:
Thank you!

PR #319, opened 2019-02-25
Funspecsubsume

lennartberinger:
Support for subsumption of funspecs


andrew-appel:
Support for funspec subsumption (subsume_funspec renamed to funspec_sub) not only at function-call sites

PR #322, opened 2019-03-13
Merging IO branch

mansky1:
Several improvements to the interface for external functions in veric, resulting in the ability to prove semax_prog_correct theorems for programs with I/O and derive top-level dry correctness properties for them.


mansky1:
Ran `make all`.

PR #346, opened 2019-06-06
Iris in VST

mansky1:
Users do not need to install Iris to build any of the material in `progs` or write new VST proofs. However, by importing `veric.bi`, a user can use Iris Proof Mode tactics to prove entailments. The material on invariants and SC low-level atomics has been ported to use IPM, and verif_hashtable_atomic demonstrates the effectiveness of the tactics. The main cost of this PR is that `compcert.lib.Maps` cannot be `Import`ed in any user-level proof that uses IPM, or `Export`ed in any file that might be `Import`ed in such proofs, to avoid a notation conflict between CompCert and std++. Any file in VST that uses CompCert's maps must explicitly `Import compcert.lib.Maps`. Also, for reasons yet undetermined, in any file that uses Iris there must be a space between any open paren and the prop operator `!!` (i.e., `( !!P)` rather than `(!!P)`).

The main benefits of IPM are 1) named separating hypotheses, allowing for a greater level of control and precision (and removing the need to rewrite with `sepcon_comm` and `sepcon_assoc`), and 2) easy handling of modalities (later, ghost updates/view shifts, etc.).


mansky1:
A branch of Iris, soon to be merged into master, removes the issue with `(!!`.

andrew-appel:
Replaced by PR #457

andrew-appel:
I misunderstood.  This is not replaced by #457, it's that #457 is merged into this one.

PR #360, opened 2019-08-05
Overlay for Coq PR #9918 fixing template-poly bug

mattam82:
This fixes an hypothesis name whose sort is Type now (so gets assigned X) but was assigned H5 before (which is definitely strange given it's a subset on a computational object). This is backwards compatible.


mattam82:
This is for coq/coq#9918

PR #365, opened 2019-09-16
Avoid relying on `Export` bugs (again)

maximedenes:
It seems that a dependency on the bug was reintroduced.

Hopefully the patch can be integrated very soon in Coq, to avoid further issues.


gares:
please merge

gares:
thanks!

PR #374, opened 2019-10-23
Try increasing the timeout for the build of Coq

maximedenes:
Please do not merge, in case CI succeeds, we will adjust the timeout to a more reasonable value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install:
 - opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev || true
 - opam update
 - opam show coq
-- opam install -j ${NJOBS} -y coq=${COQ_VERSION} ${EXTRA_OPAM}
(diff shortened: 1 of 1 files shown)


maximedenes:
@andrew-appel the coq build did terminate successfully, in about 11 minutes. So the timeout value was indeed the problem. I'll let the CI run to the end, then adjust it to 20 minutes.

andrew-appel:
That step took 690 seconds, which is about 12 minutes; so therefore I'll put a 15-minute travis-wait in the script.

maximedenes:
> so therefore I'll put a 15-minute travis-wait in the script.

I'm afraid this is going to be too tight sometimes. Build time can vary greatly from one run to another on these VMs. I'd put a timeout that is about 30 minutes, just to be sure. There's little change this step is going to start diverging anyway.

maximedenes:
`travis_wait` has been added, so closing this.

PR #394, opened 2020-03-15
[ci] [travis] Have Travis use the Docker image.

ejgallego:
I was curious to see Coq's Docker CI image working on Travis, so I did
this Proof-Of-Concept; let's see how it turns out.


ejgallego:
There may be a potential build soundness soundness bug in both this setup and the previous one: coq_makefile won't store the digest of the Coq binary used for the build; this is OK when using a stable Coq version, but using the dev one that approach would fail to rebuild when Coq does change.

PR #410, opened 2020-04-29
overlay for coq/coq#12162

olaure01:
The PR coq/coq#12162 renames `Bool.leb` into `Bool.le` which is more coherent with the rest of the standard library since it has type `bool -> bool -> Prop`. This generates possible clashes with `Nat.le` or  `Peano.le` so that additional qualification is required. `omega` also sometimes gets lost.

### DIFF
--- a/progs/ghosts.v
+++ b/progs/ghosts.v
@@ -625,7 +625,7 @@ Next Obligation.
   auto.
 Defined.
 
-Global Instance max_order : PCM_order le.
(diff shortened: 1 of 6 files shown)


andrew-appel:
Incorporated into 38b5649.

olaure01:
@andrew-appel it seems the modification of `veric/juicy_extspec.v` is missing.

andrew-appel:
Fixed in latest commit.

PR #412, opened 2020-05-05
Adapt to Coq PR #12146: tactic subst now inactive on section variables with indirect dependency in goal

herbelin:
The fix is a priori backward-compatible and can be merged now.

See [coq/coq#12139](https://github.com/coq/coq/issues/12139#issuecomment-622743684) for the motivation.

I noticed a little bug in passing in the makefile rule for building `_CoqProject`: It inserts an option `-Q compcert/. compcert..` option which is not correct.

### DIFF
--- a/floyd/VSU_addmain.v
+++ b/floyd/VSU_addmain.v
@@ -676,7 +676,7 @@ eapply Build_Component (* with (fun gv => Main_InitPred gv * (InitPred_of_Canoni
       split; [ apply semax_external_FF | apply GFF].
   - apply LNR_main_CoreG.
   - apply LNR_sort_e. apply LNR_internalspecs.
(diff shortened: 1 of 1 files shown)


andrew-appel:
I'm not sure what you mean about _CoqProject not being correct.  Read BUILD_ORGANIZATION for some explanation about how VST links to compcert.

herbelin:
> I'm not sure what you mean about _CoqProject not being correct. Read BUILD_ORGANIZATION for some explanation about how VST links to compcert.

The problem was on my side. My shell rc file had a bug setting the `ARCH` environment variable to value `.`and this was causing the `-Q compcert/. compcert..` to come up (instead of `-Q compcert/i386 compcert.i386`). So, no worry, sorry.

PR #416, opened 2020-05-12
Efficiency tweaks to entailer and other parts of Floyd

andrew-appel:


lennartberinger:
When make travis is happy, can you quickly run make in progs/VSUpile,
and (not quite so quickly) make -kj6 in vst-private/abstractprotocols/memmgr
(with your personal CONFIGURE file) before pushing to git? 
Then, I'll do make in my other applications.

The entailer application I was mentioning this morning is in line 
53 of VSU_mmap0.v (which is therefore commented out at present).
The tactic is applied in line 62.

L.

----- Original Message -----

PR #417, opened 2020-05-12
Fix compilation w.r.t. Coq master.

ppedrot:
For some reason, Coq master broke VST compilation. This fixes the compilation on master, and hopefully this should be backwards compatible. I don't have enough CPU free time to check this right now, but the CI should tell us faster.

### DIFF
--- a/veric/mapsto_memory_block.v
+++ b/veric/mapsto_memory_block.v
@@ -1375,7 +1375,7 @@ Proof.
   induction N; intros; trivial. remember (size_chunk Mptr) as sz.
   replace (Z.of_nat (S N) * sz)%Z with (sz + Z.of_nat N * sz)%Z by lia.
   specialize (size_chunk_pos Mptr); intros. specialize (Z_of_nat_ge_O N); intros.
-  eapply derives_trans. apply mapsto_zeros_split; subst; try lia. apply Z.mul_nonneg_nonneg; lia.
+  eapply derives_trans. apply mapsto_zeros_split; subst; try first [lia | apply Z.mul_nonneg_nonneg; lia].
   apply andp_right.
   { clear IHN. intros m [m1 [m2 [J [[M1 _] [[M2a M2b] _]]]]]; simpl in *.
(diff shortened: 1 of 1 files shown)


lennartberinger:
The line that broke mentions lia. We only recently replaced many occurrences of omega with lia, and just earlier today observed that a nontermination of one of our tactics was caused by lia blowing up where omega hadn't, in a case where there are about 30 inequalities of (positive) numbers in the proof context. So it could be that lia is a bit more brittle than one might hope. Anyway, as your fix solves the issue I have no objections against merging it.

ejgallego:
Thanks for the observations @lennartberinger , indeed the transition from `omega` to `lia` is not as smooth as one would hope as `lia` is not a drop-in replacement for `omega` (it is usually stronger) and it needs more battle-testing.

Coq 8.12 ships some improvements that should hopefully alleviate your performance problems, don't hesitate to open a bug.

Deprecating `omega` was not an easy decision but we believe that it will be very beneficial in the medium term as we can focus on a single arithmetic solver upstream.

ppedrot:
(In any case, the compilation passes, so it's indeed backwards compatible and should be merge ASAP so as to restore our CI.)

ejgallego:
Indeed, thanks for the fix @ppedrot, we would appreciate if it can be merged.

lennartberinger:
No need for a bug report for the issue I mentioned yesterday - the situation could be resolved by clearing all the assumptions before calling lia. Another difference between lia and omega appears to be that lia sometimes seems not to unfold variables x introduced by  "set x := ..." whereas omega did. Not sure whether that's intentional or not. Anyway, I'll now accept the PR.

ppedrot:
Thanks!

ejgallego:
> Another difference between lia and omega appears to be that lia sometimes seems not to unfold variables x introduced by "set x := ..." whereas omega did. Not sure whether that's intentional or not.

I guess this depends on heuristics and these are slightly different. I recommend reporting so the maintainers can have a look and see if that's indeed the desired behavior; but it could make sense to wait for 8.12 as there have been adjustments in `lia` over that.

PR #445, opened 2020-07-28
Make install target + make compatibility for Windows

MSoegtropIMC:
This PR adds two commits from the 2.6 release branch to master. This are the only differences between the release branch and master as of now.

One commit adds a make install target. The other commit is a minor Windows compatibility patch for the make flle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *~
 # Executables and configuration
 CONFIGURE
+VST.config
 compcert/ccomp
 compcert/ccomp.byte
 compcert/ccomp.prof
--- a/Makefile
+++ b/Makefile
@@ -220,6 +220,17 @@ endif
 FLOCQ=         # this mode to use the flocq packaged with Coq or opam
 # FLOCQ=flocq  # this mode to use the flocq built into compcert
 
+# ##### Configure installation folder #####
+
+ifeq ($(ARCH),x86)
+  ifeq ($(BITSIZE),32)
+    INSTALLDIR ?= $(COQLIB)/user-contrib/VST
+  else
+    INSTALLDIR ?= $(realpath $(COQLIB)/../coq-variant/VST64/VST)
+  endif
+else
+  INSTALLDIR ?= $(realpath $(COQLIB)/../coq-variant/VST_$(ARCH)_$(BITSIZE)/VST
+endif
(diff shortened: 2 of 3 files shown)


MSoegtropIMC:
> At some point I'd like to understand why the CI checks regarding the "docker" are failing.

There is something weird going on - I can't access the logs. I restarted the job to check if it is just a technical issue.

MSoegtropIMC:
On the second try I got the log. The issue is this:
```
  The following dependencies couldn't be met:
303    - coq-vst -> coq-compcert >= 3.7~ -> coq < 8.12
304        not available because the package is pinned to version 8.12.0
```
This should be fixed by https://github.com/coq/opam-coq-archive/pull/1352. Since for CompCert this sometimes takes a while, should I add the opam packages to the VST local opam patch repo and (if not yet done so) make the docker CI use it?

MSoegtropIMC:
My CompCert 3.7 for Coq 8.12 coq-opam PR has been merged upstream, so I retry the failed 8.12 tests.

roconnor-blockstream:
This install process doesn't work when `COMPCERT=bundled`.  I guess it only works when `COMPCERT=platorm`?

MSoegtropIMC:
I guess you mean "make install" does not work. COMPCERT=bundeld is indeed intended for development scenarios while "make install" is a production operation, so this is intended.

Are there good reasons why you would like to install the bundled CompCert?

roconnor-blockstream:
Probably not any good reasons.  I'll see about replicating the patches from http://coq.io/opam/coq-compcert.3.7~coq-platform~open-source.html in order to install an open source version of compcert, and then try with `COMPCERT=inst_dir`.

MSoegtropIMC:
I think you should just install the CompCert opam package you want (either open source or not open source - there are both) and use it with COMPCERT=platform. Unless you want to do experimental synchronized modifications to VST and CompCert, this is the way to go.

roconnor-blockstream:
I'm using nix instead of opam, so I need to create a suitable package for compcert.  The nice thing about COMPCERT=bundled was that it required no work on my part to build.  However, I should be able to model a nix package for compcert on the opam package description.  (There does already exist a compcert package in nix, but it isn't opensource.)

PR #446, opened 2020-08-02
Fix 64bit installed file list

MSoegtropIMC:
The 64 bit `make install` target did not install all files in the progs64 folder because the BITSIZE parameter has not been passed to the `calc_install_files` utility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -634,7 +634,7 @@ endif
 
 PROGS64_FILES=$(V64_ORDINARY)
 
-INSTALL_FILES_SRC=$(shell util/calc_install_files $(PROGSDIR))
+INSTALL_FILES_SRC=$(shell BITSIZE=$(BITSIZE) ARCH=$(ARCH) util/calc_install_files $(PROGSDIR))
 INSTALL_FILES_VO=$(patsubst %.v,%.vo,$(INSTALL_FILES_SRC))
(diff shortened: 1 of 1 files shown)


Zimmi48:
@MSoegtropIMC Is it normal that the `release-v2.6` branch does not contain the "Fix list of installed files" commit?

andrew-appel:
I believe the release-v2.6 branch should be deleted; it was in preparation for (and debugging) the release.  The branch actually released is the v2.6 tag of the master branch.

Zimmi48:
This is the ref that is used to build VST in the installer in Coq v8.12 currently, so please don't delete it without fixing this first.

andrew-appel:
OK, so let's wait for @MSoegtropIMC to advise about this issue, and (if necessary) rationalize our branches a bit more.

MSoegtropIMC:
@Zimmi48 : sorry I concentrated too much on the Coq platform the last 2 months. The next Coq release should indeed take tag v2.6 as the current VST opam package does. In general I think we should compare what the coq platform installs and what the Windows installer does and align these. I will release the Coq platform 8.12.0 beta 1 today.

Zimmi48:
OK, I'll update the ref. I'll let you know when it's OK to delete the branch.

Zimmi48:
See coq/coq#13056.

PR #448, opened 2020-09-01
Avoid running git when the command does not exist

roconnor-blockstream:
When building from a .tar.gz release, git will not always be available.  Do not create a git_rev definition in such a case.


andrew-appel:
Perhaps it would be better, when git does not exist, to set `Definition git_rev := ""` instead of having the definition missing entirely?

roconnor-blockstream:
Done.

PR #449, opened 2020-09-02
Pass COMPCERT env vars to calc_install_files

roconnor-blockstream:
When `COMPCERT` and `COMPCERT_INST_DIR` are passed on the command line to make rather than being set as environment variables, they do not make it into the environment of `calc_install_files`.  By passing in these two environment variables `make COMPCERT=inst_dir COMPCERT_INST_DIR=... install` now functions correctly.


roconnor-blockstream:
There is some tangentially related commentary @ https://github.com/PrincetonUniversity/VST/pull/445#issuecomment-685148642

PR #431, opened 2020-06-22
Make has_ext polymorphic

liyishuai:
Fix #430

Build failing, need for help:
> COQC progs/verif_reverse.v
File "./floyd/forward.v", line 1337, characters 4-53:
Error: No matching clauses for match.
make: *** [progs/verif_reverse.vo] Error 1

PR #451, opened 2020-09-18
Has ext poly

andrew-appel:
This is a new version of pull request #431, with everything working.  It does mean, however, that any use of has_ext from main_pre must either give an explicit universe `has_ext@{main_pre.u}` or be explicitly universe-polymorphic (in this p.r., see the file `progs/verif_reverse.v`, the `Lemma setup_globals` for an example.


mansky1:
I don’t understand Coq’s universe inference at all, but I don’t see why we should need universe annotations on completely trivial uses of `has_ext` (where the `Z` argument is `unit`). I’d like to take a closer look at Yishuai’s use case and try to figure out if we can avoid annotations in the common case.
 
One thing that might impact this is that the web server project recently switched to the `expert-external` branch, in which `has_ext` takes a ghost algebra rather than just a type as an argument. I’m not sure how that will interact with universe polymorphism, and it might be relevant to the examples Yishuai is looking at. So we might want to try out universe polymorphism based on that branch before making any decisions about the best approach.

andrew-appel:
Shall we abandon this pull request (that is, close without merging) ?

mansky1:
Yes, I think so. I need to make a new PR for `expert-external`, which addresses similar issues but probably doesn't have any actual changes in common with this.

PR #460, opened 2020-11-17
Improve error messages + fix for forward_loop

MSoegtropIMC:
This PR contains 3 commits:

- A lot of messages for the `fail` tactic are not delivered to the user. If the fail is inside of a backtracking tactic like match, match goal, first, repeat, tryif, ... the effect of fail is not to abort the tactic and to deliver the error message, but to trigger backtracking. If backtracking fails, one then gets an error like "no match" rather than the message in the fail. The fix is to say `fail 100 "message"` - this will abort 100 levels of nested backtracking. The remaining level is shown and typically around 95..97 which shows that this is required. I chose 100 to be safe in case of automation around VST tactics is used.

- There was no error message in case the condition of an if cannot be evaluated (has unknown locals).

- `forward_loop` did not work in case the loop is not followed by an instruction, e.g. a loop in an if branch without following instructions. The patch is not nice (cut & paste) but works.

### DIFF
--- a/floyd/VSU_addmain.v
+++ b/floyd/VSU_addmain.v
@@ -803,16 +803,16 @@ Ltac solve_cenvcs_goal_eq :=
 
 Ltac prove_linked_semax_prog :=
  split3; [ | | split3; [ | | split]];
- [ reflexivity || fail "duplicate identifier in prog_defs"
- | reflexivity || fail "unaligned initializer"
- | solve [solve_cenvcs_goal_eq || solve_cenvcs_goal || fail "comp_specs not equal"]
+ [ reflexivity || fail 100 "duplicate identifier in prog_defs"
+ | reflexivity || fail 100 "unaligned initializer"
+ | solve [solve_cenvcs_goal_eq || solve_cenvcs_goal || fail 100 "comp_specs not equal"]
  |
- | reflexivity || fail "match_globvars failed"
+ | reflexivity || fail 100 "match_globvars failed"
  | (*match goal with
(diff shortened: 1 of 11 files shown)


MSoegtropIMC:
There seems to be one case where a fail should be not fatal (or code relies on that it is not - let me check this).

MSoegtropIMC:
@andrew-appel : changing "fail" to "fail 100" impacts 3 of the verif_files in master. Two of them use the "step" tactic which obviously fails since it tries tactics which are not applicable and throw fatal errors then. The other one would be easy to fix (support map2 in some list length computation).

In general I think it would be preferable to make sure users get reasonable error messages. In order to also support higher level automation like `step` I see the following solutions:

- have a `fail_fatal` tactic which is usually `fail 100` but can be redefined to just `fail` for projects using higher level automation like step - this would be least invasive

- have an `inner` variant of tactics which allow to give the fail argument and have the usual tactics pass 100 for this while e.g. step could use the `inner` variant with argument 1

- use e.g. 10 instead of 100 (the maximum depth I have seen in practice is 7) and use 10 nested tries in step (wrapped in a tactic called ignore_fatal_errors)

- use a `match goal` instead of a `first` in `step` to make sure only tactics are called which make sense in the current situation - this likely needs to be combined with a light form of the above `inner` suggestion.

Alternatively one could also remove the step tactic. It is used only rarely and anyway quite slow and not so instructive.

I would be happy to implement any of these - or other suggestions you have.

Just removing the fail 100 in cases where it collides with step is I think not a good solution. It would involve many cases which are obviously intended to be fatal.

andrew-appel:
See #462

PR #462, opened 2020-11-25
Improve error messages + fix for forward_loop

andrew-appel:
This is a continuation of P.R. #460


andrew-appel:
In my changes just now to the original PR #460, I adjusted as follows:

1. list_solve does not "fail 100", so that users can do `try list_solve`.

2. The forward tactics have a "fail softly" mode, enabled by putting `FS := mk_FailSoftly`  above the line before calling forward, forward_if, etc.; and the `step` tactic uses this feature.

3. Some other tactics with "fail 100" have been adjusted.

4. The `entailer` tactic in this PR does a "fail 100" if the current proof goal is not an entailment, but it is not clear to me whether this is a good idea; perhaps we want users to be able to `try entailer!`.  

With these changes, "make travis" now builds again, but with a change in sha/verif_hmac_init.v required by point number 4 (that is, it previously did "try entailer!", and maybe we should continue to permit that).

andrew-appel:
Overall, I am still not quite sure that this "fail 100" approach is a good idea, but at least with my changes we can experiment and evaluate the concept.

andrew-appel:
error messages are suppressed inside of forward tactics. For example, see this link. The match clause eats the error message that should be exposed to the user. This is a bug introduced (I believe) when we made the "step" tactic and changed error level 100 to error level 0, without checking if error level 0 is right. The error level was set as 100 originally, to break out from all match clauses. We searched and replaced them into fail, without any further checking, and that was wrong. We should make the tactics fail with level 0 when it fails, and report error messages when proper. We do not need to set up an individual fail level for each fail, we use lazymatch. In VST-floyd, match is used in many places that lazymatch should be used, e.g. L543 in that example (entailer).

andrew-appel:
@bcip and I are still discussing whether this entire approach -- "fail 100" for forward tactics applied to inappropriate proof goals -- is a good idea.  Perhaps this PR was motivated by the fact that there were some actual bugs:  error messages were improperly suppressed inside of forward tactics.   If, for example, one puts "fail 0 foo" inside a match clause, it will be consumed by the match, and the "foo" message will never be seen.  That should be "fail 1 foo" or (in many cases simpler) it should be a lazymatch.

So an alternate proposal would be to fix all these little bugs, with the goal that "forward" will fail with a _net_ fail-level of 0, with an appropriate error message.

andrew-appel:
We will not proceed in this direction.  However, several of the fail messages have been given a more accurate fail level, so the original problem is not so severe.

PR #464, opened 2020-11-28
Simplify GitHub Action script

liyishuai:
No need for `custom_script` since coq-community/docker-coq-action#23 is resolved. Tested and ready to merge.

### DIFF
--- a/.github/workflows/coq-action.yml
+++ b/.github/workflows/coq-action.yml
@@ -26,23 +26,6 @@ jobs:
         with:
           opam_file: ${{ matrix.opam_name }}.opam
           coq_version: ${{ matrix.coq_version }}
-          custom_script: |
-            # This mostly follows https://github.com/coq-community/docker-coq-action/tree/v1.0.0
-            startGroup Opam config
-              opam repo add -y prerelease file://$WORKDIR/opam-prerelease
-              opam config list; opam repo list; opam list
-            endGroup
-            startGroup Build dependencies
-              opam pin add -n -y -k path $PACKAGE $WORKDIR
-              opam update -y
-              opam install -y -j 2 $PACKAGE --deps-only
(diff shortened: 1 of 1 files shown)


liyishuai:
Merging on Monday if no other issues arise.

.github/workflows/coq-action.yml
```diff
@@ -26,23 +26,6 @@ jobs:
         with:
           opam_file: ${{ matrix.opam_name }}.opam
           coq_version: ${{ matrix.coq_version }}
-          custom_script: |
-            # This mostly follows https://github.com/coq-community/docker-coq-action/tree/v1.0.0
-            startGroup Opam config
-              opam repo add -y prerelease file://$WORKDIR/opam-prerelease
```
liyishuai, line 32: @MSoegtropIMC removing this line didn't break the CI. If sometime you want to add it back, you can configure [`before_install`](https://github.com/coq-community/docker-coq-action#before_install) instead of the entire `custom_script`.

MSoegtropIMC:
@liyishuai : I think you should add back two things:

- the comment which connects this to https://github.com/coq-community/docker-coq-action/tree/v1.0.0 - it is a bit hard to understand how this CI script works without this link.
- The local opam patch repo should always be in CI - otherwise when one adds a patch one has to enable the repo in two different CIs. There is no good reason for not adding the empty repo. It just needlessly complicates things to add it. We have discussed the same thing before for Travis.

The CompCert patches are not required on Ubuntu, but they are required on Windows. It does make sense to express that e.g. the v2.7 tag was tested against this version of CompCert. For this reason the patches should be in (until they are upstream) and the patch repo should be active, even though the patches are not required for CI on Ubuntu. It still expresses that these patches don't break your CI.

PR #467, opened 2020-12-16
Honour the coqbench timing infrastructure flags.

ppedrot:
The coqbench uses the flag `TIMING` to output timing information to be processed further by our tools. Fortunately it's not the same as the flag `TIMINGS` already used by the makefile for the VST infrastructure...


ppedrot:
Ping.

PR #468, opened 2020-12-18
Tweak a complex unification instance.

ppedrot:
We nudge a bit unification in order to make VST compatible with coq/coq#13563. This is backwards compatible.


andrew-appel:
@ppedrot @MSoegtropIMC :  Is this required for compatibility with Coq 8.13?  If so, should I move the VST v2.7 tag to include this commit?  The idea was that VST 2.7 would work in Coq 8.13....

ppedrot:
Nope, it's only needed for 8.14 if the aforementioned PR makes it there.

PR #470, opened 2020-12-24
CI: GitHub: Add local repository back

liyishuai:
per https://github.com/PrincetonUniversity/VST/pull/464#issuecomment-750811474


MSoegtropIMC (approved):
Thanks, looks good to me!

One minor editorial note: "Print opam config" should be "Opam config", since the Opam config is also changed.

PR #471, opened 2020-12-28
OPAM: make install

### LINKED ISSUE #436: Create an "install" make target and remove the manual install in opam
Currently the VST make file does not have an install target and installation is done manually in opam and other places. From a software structure / maintainability point of view it would be better to have an install target in the VST make file and use this in opam.

liyishuai:
Resolve #436

### DIFF
--- a/coq-vst.opam
+++ b/coq-vst.opam
@@ -19,11 +19,7 @@ license: "https://raw.githubusercontent.com/PrincetonUniversity/VST/master/LICEN
 build: [make "BITSIZE=64" "-j%{jobs}%" "msl" "sepcomp" "veric" "floyd"]
 run-test: [make "BITSIZE=64" "-j%{jobs}%" "progs"]
 install: [
-	["mkdir" "-p" "%{lib}%/coq/user-contrib/VST"]
-	["cp" "-r" "msl" "%{lib}%/coq/user-contrib/VST/"]
-	["cp" "-r" "veric" "%{lib}%/coq/user-contrib/VST/"]
-	["cp" "-r" "floyd" "%{lib}%/coq/user-contrib/VST/"]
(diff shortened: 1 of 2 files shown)


MSoegtropIMC (approved):
I would say that #436 has been resolved by #67c3b1e9ab8d724466e5813c025fd0a7056f9188 and the fixes coming after it (the repo opam files use this already), so the commit message is not correct, but otherwise this commit is fine.

coq-vst.opam
```diff
@@ -19,11 +19,7 @@ license: "https://raw.githubusercontent.com/PrincetonUniversity/VST/master/LICEN
 build: [make "BITSIZE=64" "-j%{jobs}%" "msl" "sepcomp" "veric" "floyd"]
 run-test: [make "BITSIZE=64" "-j%{jobs}%" "progs"]
 install: [
-	["mkdir" "-p" "%{lib}%/coq/user-contrib/VST"]
-	["cp" "-r" "msl" "%{lib}%/coq/user-contrib/VST/"]
-	["cp" "-r" "veric" "%{lib}%/coq/user-contrib/VST/"]
-	["cp" "-r" "floyd" "%{lib}%/coq/user-contrib/VST/"]
-	["cp" "-r" "sepcomp" "%{lib}%/coq/user-contrib/VST/"]
+	make "install"
```
MSoegtropIMC, line 22: Won't you need BITSIZE=64 here?

liyishuai: Oh yes I should. I was confused by the CompCert directory logic and thought `BITSIZE=64` was the default.
https://github.com/PrincetonUniversity/VST/blob/e4211e15463cc864a06c70059c895b9257ff486a/Makefile#L80-L87

PR #475, opened 2021-01-21
Added Coq 8.13.0 to Makefile and added Imports' and associated lemma to VSU

lennartberinger:


lennartberinger:
Many Coq warnings wrt locality of hints remain.

PR #490, opened 2021-03-19
Added several lemmas on pointers & an example and fixed Makefile

zqy1018:
- Added lemmas on pointer comparisons. Excerpt:
  ```coq
  Lemma true_Cne_neq: forall x y, 
    typed_true tint (force_val (sem_cmp_pp Cne x y)) -> x <> y.
  Lemma true_Ceq_eq: forall x y, 
    typed_true tint (force_val (sem_cmp_pp Ceq x y)) -> x = y.
  Lemma false_Cne_neq: forall x y, 
    typed_false tint (force_val (sem_cmp_pp Cne x y)) -> x = y.
  Lemma false_Ceq_eq: forall x y, 
    typed_false tint (force_val (sem_cmp_pp Ceq x y)) -> x <> y.
  ```
- Added a tactic for automatically apllying the lemmas and made it one part of `forward_if`.
- Added an example (`ptr_cmp`). Excerpt:
  ```c
  struct tree {
      unsigned int k;
      struct tree *left, *right;
  };
  
  int get_branch (struct tree * p, struct tree * p_fa){
      if (p == p_fa->left){
          return 1;
      } else {
          return 0;
      }
  }
  ```
- Added Coq 8.13.1 to version check in `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ COQLIB=$(shell $(COQC) -where | tr -d '\r' | tr '\\' '/')
 
 # Check Coq version
 
-COQVERSION= 8.12.0 or-else 8.12.1 or-else 8.12.2 or-else 8.13+beta1 or-else 8.13.0
+COQVERSION= 8.12.0 or-else 8.12.1 or-else 8.12.2 or-else 8.13+beta1 or-else 8.13.0 or-else 8.13.1
 
 COQV=$(shell $(COQC) -v)
 ifneq ($(IGNORECOQVERSION),true)
--- a/floyd/forward.v
+++ b/floyd/forward.v
@@ -1728,6 +1728,126 @@ destruct (Int.eq i Int.zero) eqn:?; inv H1.
 apply int_eq_e in Heqb. subst; reflexivity.
 Qed.
 
+Lemma true_Cne_neq: 
+  forall x y, 
+    typed_true tint (force_val (sem_cmp_pp Cne x y)) -> x <> y.
+Proof.
+  intros. hnf in H. destruct x, y; try inversion H.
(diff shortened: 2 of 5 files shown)


andrew-appel:
I like this idea, but perhaps instead of applying your lemmas directly from forward_if'_new, you could integrate this into do_repr_inj (which is called from forward_if'_new).

zqy1018:
Fixed. Thanks for your tip.

andrew-appel:
I am still not sure about this.  Is there a way to better integrate it into the other logic of do_repr_inj, to make sure it's done in a style consistent with other similar operations?  For example, this code (already there) is handling a similar set of cases:

```
match type of H with
            | typed_true _  (force_val (sem_cmp_pp Ceq _ _)) =>
                                    apply typed_true_nullptr3 in H
            | typed_true _  (force_val (sem_cmp_pp Cne _ _)) =>
                                    apply typed_true_nullptr4 in H
            | typed_false _  (force_val (sem_cmp_pp Ceq _ _)) =>
                                    apply typed_false_nullptr3 in H
            | typed_false _  (force_val (sem_cmp_pp Cne _ _)) =>
                                    apply typed_false_nullptr4 in H
          end
```

zqy1018:
Sorry, I did not notice that part.

Maybe it is better to make typed_true(false)_nullptr3(4) corollaries of the new lemmas here (since they are more general), and modify that part in `do_repr_inj` into applying the new lemmas. Then no additional tactic will be needed.

zqy1018:
Seems that only applying the new lemmas in `do_repr_inj` will cause some problems. Now fixed.

PR #497, opened 2021-05-03
Support platforms where endianness is a parameter (ARM)

MSoegtropIMC:
On ARM platforms, the endianness is a parameter, rather than a defined value as for x86:

(https://github.com/AbsInt/CompCert/blob/1f35599abe1b82f565a9a1b1ee03f60df362f22d/arm/Archi.v#L24)
(https://github.com/AbsInt/CompCert/blob/1f35599abe1b82f565a9a1b1ee03f60df362f22d/x86_32/Archi.v#L24)

This leads in a few places to issues in reflexivity proofs, because terms don't evaluate cause of the parameter. This PR changes the proofs such that they also work when the endianness is unknown.

With this change the full VST build (including examples in progs) runs through with a CompCert configured for arm32 (tested with armv7a-eabihf).

Note: in some cases I pulled out an equality proof into an extra lemma - even though this lemma doesn't have any parameters - because the rewrite with `Tauto.if_same` was tricky inside of the larger lemma and also the rewrite requires careful evaluation, so that the `if` with the parameter is visible but not under a binder.

### DIFF
--- a/veric/initialize.v
+++ b/veric/initialize.v
@@ -349,7 +349,11 @@ Proof.
 intros; hnf; intros.
 transitivity
   (Some (decode_val chunk (list_repeat (size_chunk_nat chunk) (Byte Byte.zero)))).
-2: destruct chunk; reflexivity.
+2: destruct chunk; (
+     simpl list_repeat;
+     cbv delta [decode_val decode_int proj_bytes rev_if_be rev] iota beta zeta;
+     try rewrite Tauto.if_same;
+     reflexivity).
 apply loadbytes_load; auto.
 clear H2.
 rewrite size_chunk_conv in *.
--- a/veric/mapsto_memory_block.v
+++ b/veric/mapsto_memory_block.v
@@ -353,7 +353,9 @@ Proof.
  apply identity_unit'; auto.
 + exists (Byte Byte.zero :: nil); split.
  split. split. reflexivity. split.
- unfold decode_val. simpl. f_equal.
+ unfold decode_val. simpl. apply f_equal.
+ unfold decode_int, rev_if_be.
+ rewrite Tauto.if_same; reflexivity.
  apply Z.divide_1_l.
  intro loc. hnf. if_tac. exists H0.
  destruct loc as [b' i']. destruct H8; subst b'.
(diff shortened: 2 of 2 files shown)


andrew-appel:
I am not sure this is the right approach.  

- Because our CI operates on x86 where endianness is a constant, these new lemmas (or new proofs of old lemmas) will not be fully tested (in the future) by the CI, so they may be fragile -- in the sense that they will keep working for all the platforms in which endianness is a constant, but not on arm where it is a parameter.   One solution to that problem would be to add arm to the CI.
- C programs are not necessarily portable w.r.t. endianness, and with endianness as a parameter it will now be impossible to prove those nonportable programs correct, unless the user adds an axiom saying what is the endianness of their target machine.
- This will conflict with commit 1c19f9a0044ec1a85a23bff38b3bb2352b2e914a that (coincidentally) I just pushed in PR495; that commit fills in some proofs that were admitted a few months ago, regarding address_mapsto_zeros.  Proofs of lemma `address_mapsto_zeros'_address_mapsto` is likely to break.  Fixing it should not be too difficult.

But the alternative would require to have two different ARM targets in CompCert, with endianness being a constant (not a parameter) in each one.   I am not sure whether that is a realistic alternative.  Therefore I will merge this pull request, then fix up my proof of `address_mapsto_zeros'_address_mapsto` in the other pull request.

MSoegtropIMC:
For discussing the problem of verifying endianness dependent code I created (https://github.com/AbsInt/CompCert/issues/396)

Regarding CI: there are a few options:
- I plan to check this in Coq Platform CI soon, so at least one would know if it breaks within a day
- one could add a daily or weekly CI to VST which runs more extensive tests
- in the per commit CI do a mix of say 32 bit ARM and 64 bit x86 to increase coverage of variants without increasing runtime.

andrew-appel:
My proof that breaks, it turns out, is not `address_mapsto_zeros'_address_mapsto` , it is `by_value_data_at_rec_zero_val` in `floyd/data_at_rec_lemmas.v`.  Easily fixed in principle ; not as easy in practice, because my installation has `Archi.bigendian:=false`, so `decode_val` simplifies all the way, whereas it would not simplify with `Parameter Archi.bigendian`.

andrew-appel:
Commit https://github.com/PrincetonUniversity/VST/pull/495/commits/1dc4af0f1c915b4a3dc129b9b05730fac03a5cc9 attempts to fix `by_value_data_at_rec_zero_val` so it is independent of Archi.big_endian.

Issue #485, opened 2021-02-23
Make: IGNORECOQVERSION + BITSIZE=64 does not work with make install


MSoegtropIMC:
If one calls
```
make install BITSIZE=64 IGNORECOQVERSION=true
```
only a small subset of files is installed, notably no .vo files. Documentation files are installed.

MSoegtropIMC:
Fixed by #504.

PR #504, opened 2021-06-15
Fix issue #485: make install with IGNORECOQVERSION

MSoegtropIMC:
This PR fixes issue #485.

The root cause was that IGNORECOQVERSION was not passed on to sub make calls in `calc_install_files`, so that the sub make aborts immediately in case the IGNORECOQVERSION=true is actually required as in a .dev build.

I will merge this immediately since it is trivial to undo in case this should be required and this keeps me from making progress with the new Coq Platform release (including VST 2.8).

### DIFF
--- a/util/calc_install_files
+++ b/util/calc_install_files
@@ -1,5 +1,5 @@
 #!/bin/bash
 #  The $1 argument of this script should be $(PROGSDIR)
 make depend >& /dev/null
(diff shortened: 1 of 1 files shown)


MSoegtropIMC:
@andrew-appel : the Travis CI on master failed for this, but it seems to be a network issue - unrelated to my change. Can you please rerun the travis job?

PR #506, opened 2021-07-01
change size of the lock type to be 6 Tpointers wide

rinshankaihou:
The size of the lock type should be 6 pointers wide on 32-bit machines. IIRC the size is 8 Tpointers wide on 64-bit machines, but since this branch mainly serves deepweb which depends on the 32-bit clightgen, I ignore the 64-bit version on this branch.

### DIFF
--- a/concurrency/lksize.v
+++ b/concurrency/lksize.v
@@ -3,7 +3,7 @@ Require Import compcert.common.Memdata.
 Require Import Coq.ZArith.ZArith.
 
 (* LKSIZE should match sizeof(semax_conc.tlock).  *)
-Definition LKSIZE:= (2 * size_chunk Mptr)%Z.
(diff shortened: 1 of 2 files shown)


andrew-appel:
Remark 1:  Why not just define this way so it works on 32-bit or 64-bit machines:
`Definition LKsize := ((if Archi.ptr64 then 8 else 6)  * size_chunk Mptr)%Z.
`

Remark 2:  We should perhaps resolve issue #518; that is, instead of using "lock" we should be using "PThreads semaphore".

andrew-appel:
I believe that @txyyss and @mansky1 are implementing the simple spinlock as described just above.   If that is the case, can you put that into a pull-request somewhere and (at the same time) close this PR?

PR #519, opened 2021-09-23
Add msl and sepcomp to calc_install_files

intoverflow:
# tl;dr

This PR provides a quick-fix for a simple omission in `util/calc_install_files`. This omission impacts the VST installation process: in particular, it prevents `opam install` and `make install` from providing a complete installation of MSL and sepcomp theories.

# Full explanation

`Makefile` uses a shell script (`util/calc_install_files`) to determine which files to install:

```Makefile
# https://github.com/PrincetonUniversity/VST/blob/master/Makefile#L638
INSTALL_FILES_SRC=$(shell COMPCERT=$(COMPCERT) COMPCERT_INST_DIR=$(COMPCERT_INST_DIR) BITSIZE=$(BITSIZE) ARCH=$(ARCH) util/calc_install_files $(PROGSDIR))

# ...

# https://github.com/PrincetonUniversity/VST/blob/master/Makefile#L737
install: VST.config
	install -d "$(INSTALLDIR)"
	for d in $(sort $(dir $(INSTALL_FILES) $(EXTRA_INSTALL_FILES))); do install -d "$(INSTALLDIR)/$$d"; done
	for f in $(INSTALL_FILES); do install -m 0644 $$f "$(INSTALLDIR)/$$(dirname $$f)"; done
	for f in $(EXTRA_INSTALL_FILES); do install -m 0644 $$f "$(INSTALLDIR)/$$(dirname $$f)"; done
```

This shell script (re)-invokes `make` and processes the output:
```bash
# https://github.com/PrincetonUniversity/VST/blob/master/util/calc_install_files#L4
make CLIGHTGEN="CLIGHTGEN" IGNORECOQVERSION=true -Bn veric floyd $1 2>/dev/null | \
 awk '/^echo COQC /{print $NF}/^CLIGHTGEN/{print $NF}'
```

As can be seen above, the shell script provides `make` with a _hard coded_ list of targets (`veric` and `floyd`).

This list does not include `msl` or `sepcomp`. As a result, copies of VST installed via `opam install` or `make install` are incomplete, containing only those portions of `msl` and `sepcomp` that are used by `veric` or `floyd`.

This PR provides a quick-fix by adding these targets to the shell script.

### DIFF
--- a/util/calc_install_files
+++ b/util/calc_install_files
@@ -1,5 +1,5 @@
 #!/bin/bash
 #  The $1 argument of this script should be $(PROGSDIR)
 make depend >& /dev/null
-make CLIGHTGEN="CLIGHTGEN" IGNORECOQVERSION=true -Bn veric floyd $1 2>/dev/null | \
+make CLIGHTGEN="CLIGHTGEN" IGNORECOQVERSION=true -Bn msl sepcomp veric floyd $1 2>/dev/null | \
  awk '/^echo COQC /{print $NF}/^CLIGHTGEN/{print $NF}'


andrew-appel:
I am not sure this is a good idea.  There's lots in msl and sepcomp that are obsolescent, not needed by VST and perhaps not needed by anyone.  What's the motivation for this?

intoverflow:
The motivation is this line in this Makefile:
```makefile
# https://github.com/Salamari/CertiGraph/blob/live/Makefile#L176
# A minimal list of files that need to built within VST for our build to work.
# Please add to this as needed.
VST_CRITICAL_FILES = \
  concurrency/conclib.v floyd/proofauto.v floyd/library.v floyd/reassoc_seq.v compcert/cfrontend/ClightBigstep.v msl/msl_direct.v msl/alg_seplog_direct.v
```

This line also mentions `concurrency/conclib.v` which I meant to also ask about (it gets imported by CertiGraph's GC proof, though I haven't yet determined if it's truly essential).

In general, if the VST source includes extra files, then projects that build against VST by including it as a Git submodule (or similar) will have access to strictly more files than projects that build against VST from opam. (I'm currently trying to create opam packages for CertiGraph, which is how I encountered this issue.)

Agree that you should probably not install obsolete files, though. Hm...

andrew-appel:
@txyyss and @mansky1  please comment on what files should be included in all VST distributions that are not already included in the veric and floyd make-targets.

mansky1:
Do we want to include support for concurrent clients in this? We have the `concurrency` target for the soundness proofs, but we don't have a make target for client-facing concurrency rules yet (right now, those files only get built if we try to compile verif_incr or another concurrency example). This would include concurrency/lksize, semax_conc_pred, semax_conc, and then possibly also ghosts, invariants, fupd, as well as conclib. conclib is also used by some non-concurrent examples because it's just full of useful lemmas that we haven't split up and moved elsewhere (they probably belong in floyd, but I haven't had the time yet and probably won't anytime soon). So even if we don't want to install concurrency files by default, we might still want to install conclib.

andrew-appel:
I think it's fine to include concurrency/lksize, semax_conc_pred, semax_conc, ghosts, invariants, fupd, as well as conclib.   Those aren't "obsolescent" and they may be useful to users of VST.

txyyss:
Although CertiGraph's GC proof is for a sequential GC, it imports `concurrency/conclib.v` because some lemmas in it such as `data_at__eq: forall (sh : Share.t) (t : type) (p : val), data_at_ sh t p = data_at sh t (default_val t) p` are used in the GC proof. These lemmas are not about concurrency. They could be moved to some more proper place.

andrew-appel:
Let's include concurrency/lksize, semax_conc_pred, semax_conc, ghosts, invariants, fupd, conclib.
It is true that conclib has useful lemmas in it (about data_at) that are not concurrency related, and that someday in the future those should be split out, but I don't want to split them out right now because it would clash with some work I'm doing in the [bitfields](https://github.com/PrincetonUniversity/VST/tree/bitfields) branch.

andrew-appel:
This PR is subsumed by other developments, culminating in a change to calc_install_files in PR #543

PR #521, opened 2021-09-25
Make pointer_val_val a still right definition under 64-bit.

txyyss:
The CertiGraph project uses`pointer_val` and related functions such as `pointer_val_val` extensively (although the GC proof does not use it). The null pointer is different under 32-bit and 64-bit. So I made the pull request.


andrew-appel:
Could perhaps use "nullval" which is a Definition for that same "if Archi.ptr64" expression.

txyyss:
> Could perhaps use "nullval" which is a Definition for that same "if Archi.ptr64" expression.

That's right. I'm making the change. Should I close this PR and issue a new one?

andrew-appel:
You don't need to close this PR.  You can push a new commit to the branch, and that will update this PR.

PR #522, opened 2021-10-01
Adapt to CompCert's new bitfield represention in struct/union

andrew-appel:
CompCert [pull request 400](https://github.com/AbsInt/CompCert/pull/400) adds a new feature to the Clight ASTs:  fields of structs or unions can be bitfields.  Thus the "member" type is no longer just (ident*type) , it is now an inductive datatype with Member_plain and Member_bitfield.  This PR adds support for the new member type to VST.    Programs that use bitfields are not supported in VST, only "plain" bitfields are supported.

In addition, this PR supports a more recent commit to the CompCert master branch that moves compcert/exportclight/* to compcert/export/*.


andrew-appel:
Do not merge this P.R. until the next version of CompCert is released, or nearly released.  This P.R. is incompatible with the released version of CompCert.

andrew-appel:
Although the continuous integration fails (either a configuration error, or lack of access to CompCert master branch, or both),  this pull request seems to work perfectly (in both 64-bit mode and 32-bit mode) on my own machine, in Coq 8.13.2 with the current CompCert master branch, which should be similar to CompCert 3.10.   I consider it "ready to merge".

PR #525, opened 2021-10-09
Update sha development to work with both 32-bit and 64-bit CompCert.

roconnor-blockstream:
This patch allows `make sha` to run successfully when building with both 32-bit and 64-bit versions.


andrew-appel:
Upon further investigation, this P.R. seems rather incomplete.  In particular, in sha.c there are several instances of parameters and local variables of type `size_t`, such as `size_t n` and `size_t len`.  In the funspecs (in spec_sha.v) these would have to be Vptrofs instead of Vint.

PR #533, opened 2021-11-23
Fix CI

SkySkimmer:


SkySkimmer:
(hopefully)

andrew-appel:
Did we intend to use COMCPERT=bundled in VST's own CI?  I discussed the idea that Coq's CI for VST used "bundled", but I believe we intend that VST's own CI should get CompCert from opam.

SkySkimmer:
compcert 3.10 doesn't have a opam package so not sure what the alternative is

Zimmi48:
The remaining issue, which is specific to 8.14 and the 32bit version, seems like a real one:
```
  - File "./sha/general_lemmas.v", line 56, characters 34-53:
  Error: - Error: Found no subterm matching "?M3114 mod ?M3115" in the current goal.
```

As for getting CompCert 3.10 from opam, I'd rather defer it to @MSoegtropIMC to add the package, since he has done it in the past and best knows what to do.

andrew-appel:
In other PRs, some of which are already in the master branch, we are fixing the CI in a different way.

PR #535, opened 2021-12-20
Change the way submodules are fetched to fix CI.

Zimmi48:
Untested, but I think this small change might fix the CI issue.


Zimmi48:
Incidentally, this means that we have to update the Docker-Coq-Action template at https://github.com/coq-community/templates. It turns out that none of the Coq-community users of this template used the submodule feature, so we did not detect this regression.

Zimmi48:
Unfortunately, this isn't enough to fix CI, and I also discover in passing that there was already a PR attempting to fix it that was opened before :confounded: (I should have checked!).

PR #538, opened 2022-01-15
Removed submodule checkout step

intoverflow:
This PR attempts to fix the CI.

We will know if the fix works based on how the CI responds to this PR.

# What changed?

I'm not certain. However, I do know that the [docker-coq-action](https://github.com/coq-community/docker-coq-action/) is designed to install from the local opam file, which maybe (maybe?) does not require an explicit submodule checkout step.

We'll see.

### DIFF
--- a/.github/workflows/coq-action.yml
+++ b/.github/workflows/coq-action.yml
@@ -26,8 +26,6 @@ jobs:
           - coq-vst-32
     steps:
       - uses: actions/checkout@v2
-      - name: Checkout submodules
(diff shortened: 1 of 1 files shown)


intoverflow:
I spoke with @andrew-appel who caught me up on the issue. It is not as simple as "don't checkout the submodules." The problem relates to the (non) availability of bleeding-edge dependencies in the opam repositories. The fix, therefore, will require either (a) for the upstream dependencies to publish new versions in opam or (b) to find another way to support bleeding-edge versions.

PR #542, opened 2022-02-04
Adapt 'make all' testcases to Coq 8.15

andrew-appel:
Many fixes to test cases in sha,hmacdrbg,aes,mailbox,progs,tweetnacl, etc. so that now everything builds in Coq 8.15 and (probably) still builds in Coq 8.13.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -678,19 +678,24 @@ endif
 
 # ########## Targets ##########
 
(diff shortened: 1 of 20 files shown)


mansky1 (approved):
Weird that division changed! Also weird that the size of global variables changed in CompCert. But I don't see any problems with anything.

PR #544, opened 2022-02-14
Locality annotations

andrew-appel:
To address deprecation warnings about Hint Rewrite without locality
attribute, every Hint Rewrite now has #[export] locality,
except for a very few that were in a Section and need Local,
and a very few that were in a Module and need Global.
"make all" works properly in 8.15, in this commit.
Still many deprecation warnings remaining about Instances.


andrew-appel:
Unfortunately, this branch cannot build in Coq 8.13.   Coq 8.13 does not permit an #[export] attribute on an Instance.  Coq 8.15 deprecates any instance without an attribute. 

My intention with this P.R. was to avoid Global things as much as possible, and attach the #[export] attribute (when possible) instead of Global.  But in doing so, I cannot keep compatibility with Coq 8.13.  That's really annoying.

andrew-appel:
This branch now works in both Coq 8.13 and Coq 8.15.   There are a whole bunch of Instances and Hint Rewrites that are marked with a comment, showing how they can be adjusted for better locality in Coq8.15 (and after), once we finally abandon Coq 8.13.

PR #547, opened 2022-02-28
CI: change GitHub actions to separate build and test jobs

MSoegtropIMC:
Furthermore use make targets rather than opam files

This is work in progress.

### DIFF
--- a/.github/workflows/coq-action.yml
+++ b/.github/workflows/coq-action.yml
@@ -18,26 +18,103 @@ jobs:
       matrix:
         coq_version:
           # See https://github.com/coq-community/docker-coq/wiki for supported images
-          - '8.13'
-          - '8.14'
+          # - '8.13' Disabled for testing
+          # - '8.14' Disabled for testing
           - '8.15'
-        opam_name:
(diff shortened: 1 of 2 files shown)


intoverflow:
Is it possible to do Build Phase using the opam files, then Test Phase using Make?

MSoegtropIMC:
@intoverflow : one could probably do both with opam files, but @andrew-appel wanted to have more flexibility in this, so we agreed to switch to make targets.

MSoegtropIMC:
@andrew-appel : this is ready for review. I enabled those tests which actually pass - but for performance reasons not in all build variants.

One interesting point: the "test" target runs 70 minutes on 32 bit and <15 minutes on 64 bit. I guess the tests do not the same in both cases. You might want to refactor this so that the extra things "test" does on 32 bit go into a separate target, so that one can more easily gather statistics over run time for 32 and 64 bit.

andrew-appel:
Master-branch VST now has "test2" building correctly in 64-bit mode.
Now I'll merge this PR into master.

PR #549, opened 2022-03-06
Nontrivial Iris-style persistency

mansky1:
This PR lets us treat nontrivial ghost state, including invariants, as persistent resources in IPM (e.g., we can use the `#` intro pattern with them). This saves us from having to manually duplicate and deallocate invariants, and also lets us use Iris's definition of atomic updates directly. As a side effect, ghost state in general can now be deallocated without a view shift.

This required two changes to the foundations of VST. First, the `join_core` axiom from `msl/sepalg.v` has been weakened: if two elements of an algebra are related then their cores are also related, but need not be equal. This allows for the existence of nontrivial cores, like the knowledge that a piece of state has reached at least some value. Second, all ghost state has been made affine: `emp` means that an rmap's non-ghost resources are identities, but it may still have nontrivial ghost state. This makes the persistent modality absorbing as required by Iris, and also means that our logic is now *semi-affine*: memory is malloc-free but ghost state is garbage-collected. Somewhat surprisingly, this requires almost no changes to the existing rules of Verifiable C. (Details: I removed some of the rules for `ewand`, but they could probably be restored by taking another look at its definition, and anyway I don't think we ever use it. Also, the relationship between later and imp is a bit weaker, but there were workarounds in every place it mattered.)

### DIFF
--- a/atomics/general_atomics.v
+++ b/atomics/general_atomics.v
@@ -8,9 +8,8 @@ Require Import VST.floyd.library.
 Require Import VST.floyd.sublist.
 (* trying to import Iris's definition directly *)
 Require Export iris.bi.lib.atomic.
+Require Import Program.Equality.
 
-
-Set Bullet Behavior "Strict Subproofs".
 
 (* Thoughts on invariants and the two-level structure:
    I expect that our version of the operational semantics will keep nonatomic locations as is,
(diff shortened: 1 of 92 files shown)


txyyss:
Do I still need to redefine `funspec_sub` after merging?

mansky1:
Yes. This doesn't yet integrate fancy updates into the juicy semantics, which is what will be required to allow us to open and close invariants in `funspec_sub` proofs. (Actually, we'd have to redefine `funspec_sub` even then, but that's what it would take to prove the new definition sound. :) )

andrew-appel:
Note to @mansky1: I see that this was merged into a branch that has not yet become a pull request.  When you do turn atomic_locks into a pull request, please make sure the github actions are up to date as described in the second comment by @MSoegtropIMC in issue #550.  In fact, it's probably worth turning atomic_locks into a Draft Pull Request now, so you can test this earlier rather than later.

mansky1:
Okay, I've set it up now. I think the merge got the new actions automatically, but if not, I'll debug it based on that issue.

PR #555, opened 2022-03-15
Iris-style persistency and atomic locks

mansky1:
This PR lets us treat nontrivial ghost state, including invariants, as persistent resources in IPM (e.g., we can use the `#` intro pattern with them). This saves us from having to manually duplicate and deallocate invariants, and also lets us use Iris's definition of atomic updates directly. As a side effect, ghost state in general can now be deallocated without a view shift. It also includes a spinlock implemented with C11 atomics, which should become our new canonical lock implementation in a future PR.

This PR includes two changes to the foundations of VST. First, the join_core axiom from msl/sepalg.v has been weakened: if two elements of an algebra are related then their cores are also related, but need not be equal. This allows for the existence of nontrivial cores, like the knowledge that a piece of state has reached at least some value. Second, all ghost state has been made affine: emp means that an rmap's non-ghost resources are identities, but it may still have nontrivial ghost state. This makes the persistent modality absorbing as required by Iris, and also means that our logic is now semi-affine: memory is malloc-free but ghost state is garbage-collected. Somewhat surprisingly, this requires almost no changes to the existing rules of Verifiable C. (Details: I removed some of the rules for ewand, but they could probably be restored by taking another look at its definition, and anyway I don't think we ever use it. Also, the relationship between later and imp is a bit weaker, but there were workarounds in every place it mattered.)


andrew-appel:
I am just about ready to merge this PR.  Has someone already tested the whole thing with "make all" in 32-bit mode?  (That includes some test cases that are not part of our standard CI suite.)

Also, the following files have some warning messages about instance locality.  Please fix them.  (I am willing to merge the pull request first, if you promise to fix them later.)

- msl/knot_full_sa.v, line 82, instance locality warnings
- msl/knot_full_variant.v, lines 83, 942, 1006 instance locality warnings
- msl/knot_hered.v, lines 692, 700
- msl/knot_shims.v, line 333
- msl/predicates_hered.v, lines 45, 88
- msl/sepalg.v, lines 160, 169, 856, 863, 872 

The official way to fix them is:
1.  If possible, use the #[export] tag.
2.  Or, if you are sure that this must be Global, use the Global tag.
3.  Or, if you can't get it to work with #[export], but you're not sure it must be Global, use the #[global] tag.
Of course, #[global] is semantically the same as Global, but in the VST development it indicates that someday we may come back to this and try to make it non-global.

txyyss:
I just fixed the locality warnings, including files mentioned above.

andrew-appel:
Oops... forgot to mention that we want to maintain compatibility with Coq 8.13, which does not yet support #export locality in Instances.  So we write it like this:
`#[(*export, after Coq 8.13*)global] `

Also, the CI is failing in the 8.15 build: there's a coq-version-number problem in the makefile;  this may have been fixed already in the master branch, do you need to merge master?

txyyss:
I see. Let me fix the `#[export]` problem and merge master to the branch back.

txyyss:
Now it passed all the tests.

PR #563, opened 2022-04-16
Unfolding in forward

ZippeyKeys12:
If a forward would fail in `sc_set_load_store.load_tac`'s `hint_msg`, we will now try to find an appropriate definition and unfold it.

### How it works:
Using a similar lemma to `hint_msg_lemma`, we get the pointer, `p` which we need in a `data_at _ _ _ p` or similar for.
We then check if one exists, if not, we check if any `mpred` has `p` as an argument, if so, unfold it, `Intros`/`Intros x` and check if it now contains a `data_at _ _ _ p` or similar.

### Limitations:
Currently, it is only looking at the base pointer, so if we get `offset_val 4 p`, we'll only look for `p`.
This means, we won't try to unfold to get a `field_at` if another one exists, even if it's the wrong member.

### Questions:
Currently, I removed `unfold`'s in `verif_bst.v` that are now done automatically, as a test.
Is this fine as a test?

### DIFF
--- a/floyd/forward.v
+++ b/floyd/forward.v
@@ -3884,6 +3884,7 @@ Ltac forward :=
            try_forward_store_union_hack e1 s2 id1 t1
     | |- semax _ _ (Ssequence ?c _) _ =>
       check_precondition;
+      check_unfold_mpred_for_at;
       eapply semax_seq';
       [ forward1 c
       | fwd_result;
--- a/floyd/sc_set_load_store.v
+++ b/floyd/sc_set_load_store.v
@@ -530,7 +530,7 @@ Ltac solve_msubst_efield_denote :=
              rewrite (Ptrofs.to_int64_of_int64 (eq_refl _));
              reflexivity
             | |- Vptrofs _ = Vptrofs _ => reflexivity
-            | |- ?A = _ => fail 99 "Your subscript expression evaluates to" A "which is not in the form (Vint _) or (Vptrofs _).  Perhaps rewrite one of your LOCAL(temp _) clauses."
+            | |- ?A = _ => fail 4 "Your subscript expression evaluates to" A "which is not in the form (Vint _) or (Vptrofs _).  Perhaps rewrite one of your LOCAL(temp _) clauses."
            end
         ]
       | solve_Ptrofs_eqm_unsigned
@@ -839,6 +839,10 @@ Ltac find_unfold_mpred_aux P A p :=
   handle_intros;
   has_at_already p.
 
+(* This tactic is needed for matches where `p` is
+  not the last argument, since we have the entire mpred
+  expression. Note the two places it is called.
+*)
 Ltac find_unfold_mpred_aux' P A p :=
   lazymatch A with
   | ?A' _ _ _ _ => find_unfold_mpred_aux' P A' p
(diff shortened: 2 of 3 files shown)


andrew-appel:
Thank you. It seems like the tactics `find_unfold_mpred_aux2` and `find_unfold_mpred_aux'` could be combined into a single tactic that does not have so many clauses, something like,

```
Ltac find_unfold_mpred_aux2 R1 P p :=
 lazymatch R1 with
    | data_at => fail
    | field_at => fail
    | data_at_ => fail
    | field_at_ => fail
    | memory_block => fail
    | ?R _ => find_unfold_mpred_aux2 R P p
    | _ => find_unfold_mpred_aux P R p
 end.
```

(Warning:  I wrote these lines in the github conversation window, haven't tested them at all)

andrew-appel:
Also:  note failure of Continuous Integration:  coq 8.15, "make test", 32-bit mode.

ZippeyKeys12:
> Thank you. It seems like the tactics `find_unfold_mpred_aux2` and `find_unfold_mpred_aux'` could be combined into a single tactic that does not have so many clauses, something like,
> 
> ```
> Ltac find_unfold_mpred_aux2 R1 P p :=
>  lazymatch R1 with
>     | data_at => fail
>     | field_at => fail
>     | data_at_ => fail
>     | field_at_ => fail
>     | memory_block => fail
>     | ?R _ => find_unfold_mpred_aux2 R P p
>     | _ => find_unfold_mpred_aux P R p
>  end.
> ```
> 
> (Warning: I wrote these lines in the github conversation window, haven't tested them at all)

`find_unfold_mpred_aux'` is separate to recursively unfold representation predicates that don't have `p` as the last argument. It can be eliminated if we only unfold when `p` is the last argument. I wasn't sure about the extra matching versus being less general, so I left it as-is. If we'd rather go with only unfolding those with `p` as the last argument, I can remove `find_unfold_mpred_aux'` altogether.


> Also: note failure of Continuous Integration: coq 8.15, "make test", 32-bit mode.

Sorry about that, I've been trying to replicate it since it failed.

andrew-appel:
OK, thanks for the explanation about the last-argument issue.  Maybe explain that in a comment in the  .v file?

Just in case the CI failure is spurious in some way, I can just rerun that one.

ZippeyKeys12:
The issue was with a `fail 99` in `solve_msubst_efield_denote`. I've reduced the failure level to 4, which means it leaves `solve_msubst_efield_denote` at `0`.

Looking at the uses of `solve_msubst_efield_denote`, most of them are followed by a `|| fail 1000`, so failure level doesn't matter for them. For the remaining 3, it is first `check_unfold_mpred_for_at_aux`, from this PR. Then, `hint_msg`, which as the last element of a `first`, should fail out the same. 

Finally, there's `SEP_type_contradict`, which is only used in `first`'s followed by `hint_msg`. Now it won't fail quite as quickly, it'll fail in `hint_msg`. So this should be the only shift in functionality from the failure level change.

If the other `fail 99` in `solve_msubst_efield_denote` shows up on CI, I'll adjust it too, just don't have an example of it yet.

andrew-appel:
You can replicate the CI workflow in your own clone of VST, on your own computer.  Every test is just a `make` target that's available in the top-level makefile, and you can narrow it down to one file by reading the CI failure output and seeing which .v file it was, and doing `make progs/whatever.vo` in your own clone.

PR #572, opened 2022-04-28
Tactics: `Intros *`, `Intros ?` and `Exists ?`

ZippeyKeys12:
## `Intros ?`

Matches on the `EX ?x _, _` to do `Intros x'`, where `x' = fresh x`

Unlikely to really be used in manual proofs, it's primarily intended for automation, so now `step` doesn't just introduce `x`, `x0`, `x1`, etc. Instead it has appropriate names.

However, just in case, allows using `?` with up to two arguments, like `Intros p ?.`, `Intros ? p.`, or `Intros ? ?.`
Don't think higher arities are worth supporting.

### Why not do what `Intro''` does, but with `intro` instead of `intro a`?

If we have `EX i : _, _` and try `Intros ?`, it'll make a fresh version of `x` instead of `i`

## `Intros *`

Effectively, `(repeat Intros ?); Intros.`.

## `Exists ?`

For automation, it does a similar thing as `Intros ?`, but with `Exists`.
Useful for entailments of the form: `EX x : T, P |-- EX x : T, Q`, where you end up doing `Intros x. Exists x`

Supports up to two arguments, to mirror the `Intros ?` notation.


andrew-appel:
Can you explain what  "Intros _"  does?  By analogy with "intros _",  I would guess that it introduces a value and then discards it.  But perhaps that is not what it does.

Also, when you add a feature like this, it's a good idea to describe it in the reference manual,  in doc/VC.tex and doc/VC.pdf.

ZippeyKeys12:
It infers the name to use `Intros x` with, where `x` is taken from the `EX x`, so `EX p : val, G p` introduces `p`, `p0` or `p1`, etc.
It is true though, it does look like it would discard the value. Usually omitting a name is the style, `intros` instead of `intros a b c`, but we can't because that's not how `Intros` is set up.

What about `Intros %`? Iris uses it in `iDestruct` for a similar purpose.

andrew-appel:
Can you be guided by the syntax for "intros" ?  
([reference manual](https://coq.github.io/doc/v8.15/refman/proof-engine/tactics.html?highlight=intros#grammar-token-intropattern))

In particular:
- intros ?        (introduce one identifier, let Coq choose the name)
- intros *     (introduce all the variables, but not  ->  hypotheses, let Coq choose name)
- intros **    (introduce all variables and hypotheses,  let Coq choose names)

So, which of these is your new feature most similar to?

mansky1:
@lennartberinger , I think this `Exists _` is just automatically finding something with the same name as the binder and instantiating with it. What you want is probably a variant of `EExists` that lets you write something like `EExists (existT myT _)`.

ZippeyKeys12:
> So, which of these is your new feature most similar to?

I've changed the notation to be clearer now from `Intros _` to `Intros ?`.

Not sure if it should be `Intros *` or `Intros **`, since it does add hypotheses.
`Intros *` could be without `finish_Intros` and `Intros **` with it?

PR #576, opened 2022-05-10
Adapt w.r.t coq/coq#16004

Alizter:
If I've understood correctly, you are no longer using 8.13 so this shouldn't pose any issues.


QinshiWang:
Actually, ProD3 is still using Zlist with Coq 8.13. Maybe we should update as well.

PR #591, opened 2022-06-22
Add data_at lemmas for arrays and numbers

joscoh:
Added data_at lemmas from FEC code. The array lemmas are taken from the FEC repo (proofs/VST/CommonVST.v); the number lemmas are from a branch from a few months ago, and will be used for the proofs of the FEC actuators.

A few organizational questions:
1. Should the lemmas about `sameblock` be put somewhere else?
2. 3 lemmas (`Zlength_concat`, `combine_sublist`, and `combine_remove_nth`) as well as the `remove_nth` definition are general and not related to `data_at` specifically; should they be moved?
3. Is `iter_sepcon_arrays` of general enough interest to add to VST? Right now, it only works with byte arrays, but it provides convenient mechanisms for working with 2D byte arrays represented as an array of pointers.

### DIFF
--- a/floyd/data_at_lemmas.v
+++ b/floyd/data_at_lemmas.v
@@ -264,3 +264,1249 @@ red. simpl. normalize. rep_lia.
 rewrite !Znth_overflow by (autorewrite with sublist; auto).
 apply prop_ext; split; intros; contradiction H2; auto.
 Qed.
+
+Require Import VST.msl.iter_sepcon.
+Require Import VST.floyd.go_lower.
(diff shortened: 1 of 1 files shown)


andrew-appel:
In reviewing `Zlength_concat`, `combine_sublist`, `combine_remove_nth`, it seems to me that they are special-purpose enough (and mainly there for use in a lemma related to data at) that they might as well remain here, instead of move to the zlist library.

andrew-appel:
Replaced by pull request #592

PR #598, opened 2022-07-12
Create the catalog of examples

andrew-appel:


QinshiWang:
Joomy and I have a proof of minimal editing distance with both VST proof (Joomy) and functional correctness proof (I). It should be worth adding it here.

- [ ] Ask for Joomy's permission
- [ ] Bring it up to date

PR #601, opened 2022-07-13
Fancy updates in the model, and locks as a library

### LINKED ISSUE #518: progs64/incr.c makes invalid assumptions
When I compile and run the example program `progs64/incr.c`, which is supposed to increment a variable twice concurrently, I get a result of 1 instead.

The issue is that in pthreads, a mutex must be unlocked by the same thread that locked it, whereas in this program, the main thread locks a mutex to be unlocked by a spawned thread, which is undefined behavior according to the pthreads documentation.

mansky1:
This PR overhauls two major aspects of concurrency in VST, and is probably the last big piece of Iris-in-VST (though we still need improved tactics and documentation).

1. VST's semantics now includes fancy updates (i.e., accesses to global invariants) between steps. This allows us to access global invariants at any time in a proof using viewshift_SEP, and removes the need for some previously Admitted lemmas in `atomics/general_atomics`. Because invariants and fancy updates are in the model, they are implemented totally independently of Iris; `concurrency/invariants` and `concurrency/fupd` instantiate the relevant Iris typeclasses with VST's implementation, giving us invariants with namespaces and use of the `iInv` tactic.

Caveats:

Because of an extra step in the return operation in VST's semantics vs. CompCert's, we can't do fancy updates immediately after a function call. This means that some fancy functions (e.g., allocating a lock invariant) have to return their postconditions under a modality (that the caller can immediately remove using viewshift_SEP).

Unlike ghost-state updates, fancy updates have a major impact on the soundness proof. From a per-thread perspective, we may have hidden some of our memory permissions in an invariant, and then proved safety with the reduced permissions. To recover soundness, we need some new frame-style properties of external calls. Most of them are provable in all existing instantiations, but one in particular seems like it needs to be an axiom at the CompCert level. Specifically, we need to know that inline external calls can be run on memories with increased permissions, and ignore the increased permissions. This is a variant of CompCert's `ec_mem_extends` axiom, but with a much tighter relation than `Mem.extends`.

2. Locks as a library. This PR replaces the axiomatized Pthread synchronization primitives with implementations using C11 atomics. Our locks are now a library verified in VST. The basic lock API is changed slightly: `makelock` and `freelock` now allocate/deallocate the lock rather than operating on pre-allocated memory, and the pointer argument of `lock_inv` has been replaced by a `lock_handle` that can contain both the pointer and other information (e.g. ghost names). There are two versions of the lock implementation proofs: `atomics/verif_lock`, which is independent of Iris and provides the old lock-invariant-based specs, and `atomics/verif_lock_atomic`, which uses Iris and provides TaDA-style atomic lock specs (plus a few different invariant-based derived forms). There is also a typeclass for further lock implementations. This resolves #518. This does add proof rules for SC atomics to the TCB; proving their soundness is future work.


andrew-appel (approved):

mansky1:
@andrew-appel One last question before I merge: is it okay that this fails on 8.14? I didn't put in the compatibility hack for renamed variables in lemmas pre-8.15.

andrew-appel:
It's not ideal that it fails in Coq 8.14, but I suppose it's OK.  The goal is to maintain compatibility of every VST release with the most recent 2 versions of Coq.  Right now that's 8.14 and 8.15, but probably before the next release Coq 8.16 will be out.

PR #604, opened 2022-07-18
improved auto goal selection

mrhaandi:
Improves robustness in case of stronger (e)auto (see [coq/#16293](https://github.com/coq/coq/pull/16293)).

### DIFF
--- a/compcert/lib/Integers.v
+++ b/compcert/lib/Integers.v
@@ -1258,7 +1258,7 @@ Qed.
 
 Theorem and_zero: forall x, and x zero = zero.
 Proof.
-  bit_solve. apply andb_b_false.
+  bit_solve; apply andb_b_false.
 Qed.
 
 Corollary and_zero_l: forall x, and zero x = zero.
@@ -1268,7 +1268,7 @@ Qed.
 
 Theorem and_mone: forall x, and x mone = x.
 Proof.
-  bit_solve. apply andb_b_true.
+  bit_solve; apply andb_b_true.
 Qed.
 
 Corollary and_mone_l: forall x, and mone x = x.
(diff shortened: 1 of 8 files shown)


andrew-appel:
Is this a Coq 8.16 thing?
I am slightly concerned about the overlap between this and pull request #601.  In that branch, two or three of the files modified here are substantially modified; and we don't have a setup for testing in Coq 8.16.
  I suggest:  First merge #601 (which should happen quite soon), then merge master back into this branch and make sure everything still works.

mrhaandi:
> Is this a Coq 8.16 thing?

This is for 8.17, making `auto` slightly stronger (more close to `simple apply`).
As a result, auto solves slightly more goals and goal management becomes more important for backwards compatibility.

> I suggest: First merge #601 (which should happen quite soon), then merge master back into this branch and make sure everything still works.

Yes, this is fine.
If possible, I would like to discourage non-monotone use of `(e)auto` (see https://github.com/coq/coq/wiki/Anti-patterns#Non-monotone-automation). For example `tactic1; auto. tactic2.` breaks if auto becomes stronger (or has fewer bugs), making this PR necessary in the first place.

mrhaandi:
> In the case of `tactic1; auto. tactic2`, should it better be noticed when `tactic1` solves, in terms of proof engineering methodology?

The issue is rather `tactic2`.
Assume that `tactic1` produces five goals, such that `auto` is supposed to solve the first four and `tactic2` solves the last one. `tactic1; auto. tactic2` may work for now. However, if `auto` gets stronger then there will be no more goal left for `tactic2` to solve and it will complain. This constitutes non-monotone use of automation.
Alternatively, one could use `tactic1; [auto..|]. tactic2`. Here, regardless how strong `auto` may become, it always leaves the last goal to `tactic2`.

mansky1:
Thank you for your patience! #601 is now merged.

mrhaandi:
I rebased the PR on top of the current master after #601.

Issue #597, opened 2022-07-12
Mission of `.opam` files in this repo


QinshiWang:
I'm not sure about the mission of `.opam` files in this repo. I suppose they serve as OPAM files for the `dev` version, and prepare for the next release. But the `dev` version OPAM files are actually stored at https://github.com/coq/opam-coq-archive/tree/master/extra-dev/packages/coq-vst. The OPAM files in this repo are quite different from the released versions.

These `.opam` files don't work, either. It is because `[make "ZLIST=platform" "BITSIZE=64" "-j%{jobs}%" "vst"]` will only build `proofauto` but not examples, but `[make "ZLIST=platform" "BITSIZE=64" "install"]` will try to install some examples.

PR #606, opened 2022-07-18
Adapt w.r.t. coq/coq#16004.

ppedrot:
We probably also need to bump the submodule commits, but I'll do that in another PR.


QinshiWang:
What is
```
(** Ignore Hint Rewrite global attribute for Coq 8.13 *)
Set Warnings "-unsupported-attributes".
```
Should I add or remove some attributes?

ppedrot:
@QinshiWang VST seems to support Coq 8.13 as per the opam file, I had to keep backwards compatibility with this version. Since the Hint Rewrite locality attributes were only introduced in 8.14, unsetting the warning allows ignoring the error that Coq 8.13 raises when given global flags for Hint Rewrite commands. Once support for Coq 8.13 is dropped you can remove the Set Warning line.

QinshiWang:
@ppedrot Thanks. I believe VST has dropped support for Coq 8.13 according to the Makefile. The OPAM files in this repo are not frequent maintained (see #597). But on the other hand, the Zlist library is separately installable. It is worth supporting more Coq versions than the main body of VST.

PR #622, opened 2022-08-16
Refactor conclib

txyyss:
I removed all `conclib_*.v` files. Definitions and lemmas from these files are moved to some other existing files. The notations of dependent funspec in `conclib.v` are moved to `client_lemmas.v`.

The current branch is compilable with `make tests vst atomics` in macOS. I will test 32-bit situation and `make all` later today.


mansky1:
Thank you so much! This looks great. By the way, what happens without the `Opaque eq_dec` directives you added?

mansky1 (approved):

txyyss:
> Thank you so much! This looks great. By the way, what happens without the `Opaque eq_dec` directives you added?

Without `Opaque eq_dec`, it will be simplified by `simpl` to `EqDec_Z`. Then the tactic `destruct (eq_dec i j)` will fail.

I guess the reason is that the relocation of lemmas changes the order of dependencies of certain classes like `EqDec`.

PR #629, opened 2022-10-22
Faster tactic in a slow proof from veric/binop_lemmas4.v.

ppedrot:
We nudge unification by providing an explicit argument, leading to a 40% speedup in a very slow proof.

Hopefully this should help the Coq CI survive the stress more easily.

### DIFF
--- a/veric/binop_lemmas4.v
+++ b/veric/binop_lemmas4.v
@@ -856,10 +856,10 @@ repeat match goal with
  rewrite ?(proj2 (eqb_type_false _ _)) by auto 1;
  try reflexivity;
  try solve [simple apply test_eq_relate'; auto;
-               try (apply denote_tc_test_eq_xx; assumption);
-               try (apply denote_tc_test_eq_yy; assumption);
-               try (eapply test_eq_fiddle_signed_xx; eassumption);
-               try (eapply test_eq_fiddle_signed_yy; eassumption)];
+               try (simple apply denote_tc_test_eq_xx; assumption);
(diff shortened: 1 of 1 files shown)


andrew-appel:
The CI fails in Coq 8.14 and Coq 8.15.  Since we want to maintain two-coq-version compatibility, the failure in Coq 8.14 does not concern me, but failure in Coq 8.15 is a problem.

andrew-appel:
I will work on this, and also try to update VST's CI to do Coq 8.15,8.16 instead of its current 8.14,8.15.

ppedrot:
Sorry, this is my fault. I pushed a local version that was incorrect. This should work now.

andrew-appel:
I wonder if the same should be done to lines 864 and 865, changing these "eapply" into "simple eapply":
 
```
try (eapply test_order_fiddle_signed_xx; eassumption);
try (eapply test_order_fiddle_signed_yy; eassumption)];
```

ppedrot:
AFAICT these lines do not appear in the profile, so it's likely not to have any effect.

andrew-appel:
Is there documentation of the profiling system so I can run it myself?  Is it new in Coq 8.16 or it was there already?

ppedrot:
It's been around for a while. You can activate it using the `Set Ltac Profiling` command and display the partial result at some point of the code with `Show Ltac Profile`. There are [further options](https://coq.inria.fr/refman/proof-engine/ltac.html#profiling-ltac-tactics) to tweak this mechanism.

PR #634, opened 2022-11-09
coq-vst-iris OPAM package

mansky1:


mansky1:
It's frustratingly easy to cause an infinite makefile loop by accidentally using one of the variables that calls `util/calc_install_files` at the top level of the makefile. Especially when your editor automatically converts tabs to spaces...

PR #644, opened 2022-12-02
Change the setting of INSTALLDIR.

txyyss:
Set INSTALLDIR to the default one if ARCH is native.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,14 @@ FLOCQ=         # this mode to use the flocq packaged with Coq or opam
 
 # ##### Configure installation folder #####
 
-ifeq ($(ARCH),x86)
+MACHINE_ARCH=$(shell uname -m)
+X86_ALIASES:=x86 x86_32 x86_64 i686
+ARM_ALIASES:=arm64 aarch64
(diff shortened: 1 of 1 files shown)


andrew-appel:
This looks good to me.  I suggest you put this comment into the Makefile at line 230:
```
#  1. (if present) the VST installation for reasoning about 64-bit-mode C programs on the host 
#         architecture will be in $(COQLIB)/user-contrib/VST, 
#  2. (if present) the VST installation for reasoning about 32-bit-mode C programs for the 32-bit analogue
#       of the host architecture will be in $(COQLIB)/../coq-variant/VST32/VST
#  3. (if present) a VST installation for reasoning about C programs compiled for a _different_ 
#        architecture will be in $(COQLIB)/../coq-variant/VST_otherarch_bitsize/VST
#  Not all of this logic is right here in this makefile; some of it is done in the CompCert install,
#  in choosing how to configure CompCert itself, creating the values of $(ARCH) and $(BITSIZE)
```

MSoegtropIMC (approved):
Looks good to me, but indeed I would add the comment suggested by Andrew.

MSoegtropIMC:
Possibly one could change `IS_ARM` and `IS_X86` to `BOTH_ARM` and `BOTH_X86` since the logic is not that easy to understand.

andrew-appel:
I will go ahead and make these changes now.

PR #661, opened 2023-02-02
Use (new) bool2val instead of Val.of_bool

### LINKED ISSUE #625: Suggestion: Val.of_bool and partial evaluation
If `Val.of_bool b` were redefined as `Vint (if b then Int.one else Int.zero)` or even better `Vint (Int.repr (Z.b2z  b))`, then more partial evaluation could proceed because the constructor `Vint` is exposed.

I think this would be particularly helpful because (at least for me) `Val.of_bool` ends up occuring inside `force_val (both_int` or similar such expressions when it appears, which can reduce away as long as `Val.of_bool` can be reduced to a `Vint`.

### LINKED ISSUE #660: subsumespec_app1, subsumespec_app2
In floyd/forward.v, the lemmas subsumespec_app1 and subsumespec_app1 take forever.
Same for tycontext_sub_Gprog_app1, tycontext_sub_Gprog_app2.

Perhaps with explicitly instantiated arguments to funspec_sub_si_refl, these lemmas would be instantaneous.

andrew-appel:
This P.R. implements the suggestion in issue #625. It's an experiment, as in some cases it breaks existing VST proofs at `forward_if` and related tactics.  The changes required in those proofs are generally positive (simplifies proof scripts). See further discussion at #625.  closes #625

Incidentally and unrelated:  closes #660


andrew-appel:
x

Issue #673, opened 2023-03-30
Broken with Coq master


SkySkimmer:
eg https://gitlab.com/coq/coq/-/jobs/4025276943

andrew-appel:
I'll fix this today.   The problem is that VST's copy of CompCert is the CompCert 3.12 release, and this needs to be updated from CompCert master branch.

SkySkimmer:
BTW it seems you don't test coq master in your CI, is that intentional?

PR #675, opened 2023-03-30
Fix Coq 8.17 deprecations

andrew-appel:


andrew-appel:
closes #673 
(which is to say, it fixes at least one bug relevant to #673, but maybe there's more than one)

PR #680, opened 2023-05-04
Adapt w.r.t. coq/coq#17564.

ppedrot:
Should be backwards compatible.


ppedrot:
Ping, CI is green so it can be merged already.

PR #681, opened 2023-05-11
Detect compspecs mismatch in SEP_field_at_unify'

andrew-appel:
This fixes a bug:  during forward (load_tac), if the compspecs in the field_at did not match the current compspecs, Coq could to into a unification freeze.  Now if the compspecs are not constr_eq, you get an error message suggesting change_compspecs.


andrew-appel:
In addition, change_compspecs itself now gives a clear diagnostic message about type mismatches, rather than freezing.

PR #691, opened 2023-06-22
fix coercion of assert', fix local2ptree_denote.v

rinshankaihou:
Still no coercion from (environ->mpred) to bi_car.


floyd/local2ptree_denote.v
```diff
@@ -408,28 +409,53 @@ Proof.
...
+  P ⊣⊢ Q <-> forall rho, (P rho ⊣⊢ Q rho).
+Proof.
+  intros. split; intros.
+  - rewrite H. reflexivity.
+  - constructor. done.
+Qed.
+
+(* TODO find a better place for this *)
+(* convert assert' to assert, distribute rho among ∧. *)
+Ltac raise_rho :=
```
rinshankaihou, line 430: Is there a better place for this? It will also be used in other files that use assert'.

mansky1: Oh, yeah, I did something like that in a bunch of earlier places. The most reasonable place might be in `mpred` where it's defined in the first place.

rinshankaihou: moved to mpred.v.

PR #694, opened 2023-07-07
fix most of floyd/forward.v

rinshankaihou:
floyd/forward.v is mostly fixed, still need to fix some of its libraries.

### DIFF
--- a/floyd/diagnosis.v
+++ b/floyd/diagnosis.v
@@ -1,22 +1,22 @@
 Require Import VST.floyd.base2.
 Require Import VST.floyd.client_lemmas.
 Require Import VST.floyd.reptype_lemmas.
-Import compcert.lib.Maps.
+Import -(notations) compcert.lib.Maps.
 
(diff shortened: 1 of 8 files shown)


andrew-appel:
I suggest, in addition to the standard CI run at github, you also do "make all" locally in 32-bit mode, which runs several other test cases.  Before doing that, you may need to initialize some submodules (FCF, etc.) on which those test cases rely.

andrew-appel:
(of course, this can wait until you have made more progress on the basic test cases)

rinshankaihou:
> I suggest, in addition to the standard CI run at github, you also do "make all" locally in 32-bit mode, which runs several other test cases. Before doing that, you may need to initialize some submodules (FCF, etc.) on which those test cases rely.

I will keep that in mind, but as of now there is still some work to be done before the proof of any example (even floyd_test.c) works. By `fixed` I really mean that the file compiles but the tactics are most probably not functioning properly :)

PR #697, opened 2023-07-10
fix unification in entailer

rinshankaihou:


mansky1:
I think I figured out a different solution.

Issue #706, opened 2023-08-22
`make install` silently installs nothing when `IGNORECOMPCERTVERSION=true` is passed and the versions mismatch


JasonGross:
https://github.com/PrincetonUniversity/VST/blob/288866356f56222be47af66630602e5cc0e14490/Makefile#L671
invokes `util/calc_install_files` without setting `IGNORECOMPCERTVERSON`, and the makefile contains no `export IGNORECOMPCERTVERSION`, so https://github.com/PrincetonUniversity/VST/blob/288866356f56222be47af66630602e5cc0e14490/util/calc_install_files#L5 is run without passing along `IGNORECOMPCERTVERSION`, and the result is an error message like
```
Makefile:131: *** COMPCERT VERSION MISMATCH: COMPCERT_VERSION=version=3.12 but /home/jgross/.local64/coq/coq-master/lib/coq/user-contrib/compcert/VERSION=version=3.14.  Stop.
```
which isn't displayed due to `>& /dev/null`

PR #709, opened 2023-08-22
Pass stderr through calc_install_files when commands fail

JasonGross:
I'm not sure how good an idea this is.  It complicates the code a bit, but will probably make it easier to diagnose issues like #706 in the future


andrew-appel:
I'm not sure about this.  How about a version that does NOT run each command twice, but puts the standard error (or other info) in a log file, for later diagnosis if necessary?

JasonGross:
That's a much better idea!  I've stuck the info in `.calc_install_files.depend.log` and `.calc_install_files.make.log`, let me know if you prefer other names.

PR #714, opened 2023-09-08
Pass `COQEXTRAFLAGS` through to `coqc`

JasonGross:
This mimics the behavior of `coq_makefile` a bit more, and will allow the next release of VST to be compatible with coq-native even when its dependencies are not.  For example, compcert only installs .coq-native files since version 3.13, cf
https://github.com/AbsInt/CompCert/issues/476, and this change allows VST to be compatible with earlier versions of compcert even when coq-native is installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ endif
 # ZLIST=bundled      (default, build and use bundled zlist)
 #
 # # Choosing BITSIZE #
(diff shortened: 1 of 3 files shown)


JasonGross:
Ah, but I guess VST is specific to a particular version of compcert, so this won't be much use?

PR #731, opened 2023-10-13
removed a call to  try simple apply eq_refl in SC_tac

lennartberinger:


andrew-appel:
This is subsumed by #739

PR #739, opened 2023-10-27
Adjusted SC_tac, adjusted change_compspecs warning message, store_tac

andrew-appel:
Adjusted SC_tac so it is more careful and doesn't break things;
made the new warning message in change_compspecs user-disablable;
fixed a unification blowup in store_tac.


andrew-appel:
the commit message that refers to `load_tac` should have said `store_tac`.

andrew-appel:
@lennartberinger can you test this against your examples on which the `try simple apply eq_refl` in `SC_tac` was causing problems?

PR #781, opened 2024-07-24
Port VST 2.x to CompCert master 

andrew-appel:
https://github.com/AbsInt/CompCert/commit/f92facf8597ccd3a1af4b9f468b89d55db390bcf

which will (with future changes) become CompCert 3.15


mansky1:
Great! This looks like it won't be too bad to port -- only a few of the files will require new work on `vst_on_iris`.

For MacOS, my impression is that the developers decided they're never going to support C11 threads. We could use `#ifdef` to fall back on Pthreads when C11 threads aren't available. I don't have a Mac machine to test on at the moment, but if you have time you can try out the attached and see if it works. That said, Pthreads and C11 threads disagree on the return type of spawned functions, so this might be impossible to smooth over.
[threads.c.txt](https://github.com/user-attachments/files/16382332/threads.c.txt)

andrew-appel:
before you port to the other branch, let me get the CI working on this branch.

PR #783, opened 2024-07-26
updated mailbox to use new lock_t

### LINKED ISSUE #782: mailbox broken: typedef with/without *
The mailbox example (i.e., "make test4" or "make mailbox") is broken in the master branch, for a long time now, and the only reason we haven't noticed it in the CI is that nobody has rerun clightgen on the .c files, so we have been using very old .v files.

The reason is this:  mailbox/atomic_exchange.h includes concurrency/threads.h.
Some time ago, the typedef for lock_t was changed (in concurrency/threads.h) from `typedef struct ... lock_t;` to `typedef struct ... *lock_t;`.   But nobody updated the C code in mailbox/*.c, so they're all using  `lock_t *p`  et cetera.

Either we should update the C files in mailbox, or mailbox must have its own version of threads.h.

mansky1:
Fixes #782.

### DIFF
--- a/mailbox/atomic_exchange.c
+++ b/mailbox/atomic_exchange.c
@@ -1,7 +1,7 @@
 #include "atomic_exchange.h"
 
 //This could be replaced by an external call.
-int simulate_atomic_exchange(int *tgt, lock_t *l, int v){
+int simulate_atomic_exchange(int *tgt, lock_t l, int v){
   int x;
   acquire(l);
   x = *tgt;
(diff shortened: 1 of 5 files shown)


mansky1:
@andrew-appel I used clightgen 3.13, so please feel free to regenerate `mailbox.v` if necessary.

andrew-appel:
Instead of merging this into master, I have merged it into P.R. #781, where it works fine with CompCert 3.15.

PR #821, opened 2025-05-31
require Arith and use RelationClasses (for stdlib#150 and stdlib#152)

andres-erbsen:
Simple adaptation for https://github.com/rocq-prover/stdlib/pull/152

I am not sure why I needed to Require Arith in the other file. Hopefully that change is acceptable regardless.

Thanks!

Edit: mystery solved: `Require Arith` is for https://github.com/rocq-prover/stdlib/pull/150

### DIFF
--- a/msl/subtypes.v
+++ b/msl/subtypes.v
@@ -3,11 +3,10 @@
  *
  *)
 
+Require Import Arith.
 Require Import VST.msl.base.
 Require Import VST.msl.ageable.
 Require Import VST.msl.predicates_hered.
-
-Import Arith.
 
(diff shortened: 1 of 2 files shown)


proux01:
@andrew-appel if CI here confirms this is indeed backward compatible, could we please consider merging it so that the upstream PR on Stdlib can get merged

andrew-appel:
I can probably fix this next week.  But what is a "synchronous overlay"?

andrew-appel:
Closing without merging, because it is subsumed into #825.

PR #842, opened 2025-11-06
Fix build dependencies in Makefile; update build instruction for RefinedCC

dkxb:


Makefile
```diff
@@ -845,7 +848,7 @@ install: VST.config
 	for f in $(INSTALL_FILES); do install -m 0644 $$f "$(INSTALLDIR)/$$(dirname $$f)"; done
 	for f in $(EXTRA_INSTALL_FILES); do install -m 0644 $$f "$(INSTALLDIR)/$$(dirname $$f)"; done
 
-install_rc: VST.config refinedVST/typing/typing.vo
+install_rc: VST.config veric floyd refinedVST
```
mansky1: We shouldn't need to build all of veric or floyd for refinedVST.

rinshankaihou: I agree, I think install_rc should just depend on `$(RC_INSTALL_FILES_VO)`, like:
`install_rc: VST.config  $(RC_INSTALL_FILES_VO)`
but writing the recipe like that hangs on my machine; maybe the syntax is wrong?

rinshankaihou: This commit has been amended and dependencies is determined by the util/calc_install_files_refinedcc script now.

refinedVST/typing/automation.v
```diff
@@ -465,3 +465,18 @@ Arguments FindVal: simpl never.
...
+  we can throw P away. *)
+Ltac splitting_fast_done ::=
+    solve
+[ repeat first
+  [ fast_done
+  (* All the tactics below will introduce themselves anyway, or make no sense
+    for goals of product type. So this is a good place for us to do it. *)
+  | progress intros
+  | split ]
+| rep_lia
```
rinshankaihou: Not sure adding `rep_lia` to this is a good idea; this is intended to not introduce trivial hypotheses (otherwise we would have e.g. `0 ≤ Int.unsigned (Int.repr 1)` in the Coq context).
Also note this does not have any effect unless the proof script imports automation.v.

PR #844, opened 2025-11-12
add rules for casting to pointers

dkxb:


refinedVST/typing/own.v
```diff
@@ -322,19 +322,112 @@ Section own.
...
+    cast_int_ptr v ty = Some v'->
+    Cop.sem_cast v ty (tptr ty2) m = Some v'.
   Proof.
-    unfold int; simpl_type.
-    iIntros "HT" (Hn Φ) "HΦ".
-    iDestruct ("HT" with "[%]") as "HT".
-    { by apply: val_to_Z_in_range. }
-    iApply wp_cast_int_ptr_weak => //.
-    iIntros (i') "!>". by iApply ("HΦ" with "[] HT").
+    intros. destruct v, ty; done.
```
mansky1: I think this proof only works because under the hood it knows `Archi.ptr64 = true`. Casting an arbitrary int to `Vint 0` or long to `Vlong 0` isn't how`sem_cast` works.

mansky1:
Because cast is a separate expression type rather than a `unop`, its instances aren't used by the automation; instead, we need to apply them explicitly in `automation.v`. Right now it assumes ptr-ptr cast if the target type is pointer, and int-int cast otherwise. We could either write a combined rule for each target type that pattern-matches on the source type, or write a set of rules with fast-failing premises (like the `TCDone`s you've used here) and try them all in some order.

PR #854, opened 2026-04-20
Fix incorrect Proof commands

SkySkimmer:


ppedrot:
Ping @andrew-appel, this is backwards compatible.

PR #855, opened 2026-04-23
Adapt to https://github.com/rocq-prover/rocq/pull/21947

proux01:
Adapt to https://github.com/rocq-prover/rocq/pull/21947

Should be backward compatible

### DIFF
--- a/msl/Coqlib2.v
+++ b/msl/Coqlib2.v
@@ -5,6 +5,7 @@ Require Export VST.msl.Extensionality.
 Global Set Asymmetric Patterns.  (* This was in CompCert up to 3.9,
   and has been removed, which was a good idea but now we must have it
   here in VST until we remove it . . . *)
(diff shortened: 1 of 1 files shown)


proux01:
CI green in the upstream PR, please merge

msl/Coqlib2.v
```diff
@@ -5,6 +5,7 @@ Require Export VST.msl.Extensionality.
 Global Set Asymmetric Patterns.  (* This was in CompCert up to 3.9,
   and has been removed, which was a good idea but now we must have it
   here in VST until we remove it . . . *)
+#[warning="-unknown-option"] Global Set Asymmetric Patterns No Implicits.
```
SkySkimmer, line 8: we really shouldn't use global

andrew-appel: There are lots of Globals, some dating back to 2008 when I had no choice.  Perhaps I should make a pass through and remove as many as possible.  However, this particular P.R. makes things no worse, really.

PR #857, opened 2026-06-08
Adapt to the change in CompCert semantics between CompCert 3.7 and 3.8

andrew-appel:
The painful thing is a new restriction in CompCert's load and store semantics, that the offset must be <= Ptrofs.modulus.  This requires preserving a new invariant.


andrew-appel:
This P.R. adapts to https://github.com/AbsInt/CompCert/commit/142fc76270287ca1258d12a940c9209d22bfcc78

PR #861, opened 2026-07-16
Accept dev versions of coq, coq-compcert and coq-flocq in coq-vst.opam

JasonGross:
Allows installing a pin to this branch on a switch tracking Rocq master.

Wordsmithed by Codex.

### DIFF
--- a/coq-vst.opam
+++ b/coq-vst.opam
@@ -34,8 +34,8 @@ run-test: [
 depends: [
   "ocaml"
   "coq" {(>= "8.19" & < "9.2~") | = "dev"}
-  "coq-core" { >= "9.0" }
-  "coq-stdlib" { >= "9.0" }
+  ("coq-core" {(>= "9.0") | = "dev"} | "rocq-core" {(>= "9.0") | = "dev"})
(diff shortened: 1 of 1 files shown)


JasonGross:
> @JasonGross can you explain further?  How is this branch going to track the master branch?  Will you have to fastforward from time to time?  And why need a P.R. instead of just a fork branch?

_Originally posted by @andrew-appel in https://github.com/PrincetonUniversity/VST/issues/860#issuecomment-4993665741_
            
This is not strictly needed on master, as one can use [the version in the opam repo](https://github.com/rocq-prover/opam/blob/master/extra-dev/packages/coq-vst/coq-vst.dev/opam).  However, if we want to, e.g., `opam pin add git+https://github.com/PrincetonUniversity/VST.git#some-branch` and have that work on rocq dev, that branch would need to adjust the opam file.

The particular workflow this is useful to me for is:
1. fork VST, create a branch with a fix
2. open a PR to PrincetonUniversity/VST with the fix
3. point my opam setup at my branch in the meantime, so as to not block development of things depending on VST
4. if/when the PR is merged, repoint opam back at upstream VST

